PR #9, opened 2021-03-21
8 tag search

### LINKED ISSUE #8: Search for AWS instances by tag
Currently, the mcs-pipeline code looks for running instance types (p2.xlarge) to provision jobs. As we test the pipeline code against multiple performers/teams, it would be useful to tag EC2 instances by team and search for those tagged instances instead.

deanwetherby:
Updated utils.get_aws_machines() function to allow for EC2 instance tag search. Fixes #8 

```python
from pipeline.utils import get_aws_machines

cora_machines = get_aws_machines(tag_name="Name", tag_value="CORA")
```

### DIFF
--- a/pipeline/util.py
+++ b/pipeline/util.py
@@ -42,8 +42,8 @@ def get_s3_buckets():
 
 
 def get_aws_machines(
-        machine_type: str='p2.xlarge',
-        location: str='us-east-1',
+        instance_type: str='p2.xlarge',
+        region: str='us-east-1',
         tag_name: str='Name',
         tag_value: str=''
     ) -> List[str]:
@@ -52,7 +52,7 @@ def get_aws_machines(
     of type machine_type.
     """
 
-    ec2 = boto3.client('ec2', region_name=location)
+    ec2 = boto3.client('ec2', region_name=region)
     response = ec2.describe_instances(
         Filters=[
             {
@@ -65,7 +65,7 @@ def get_aws_machines(
             },
             {
                 'Name': 'instance-type',
-                'Values': [machine_type]
+                'Values': [instance_type]
             }
         ]
     )
--- a/pipeline/util.py
+++ b/pipeline/util.py
@@ -13,13 +13,14 @@
 # the S3 buckets and AWS machines to use.
 #
 
-
 import os
+from platform import machine
 import subprocess
 import time
 
 import boto3
 
+from typing import List
 from pipeline.secrets import Secrets
 
 PEM_FILE = Secrets['PEM_FILE']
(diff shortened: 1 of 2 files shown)


pipeline/util.py
```diff
@@ -41,36 +42,40 @@ def get_s3_buckets():
...
-            # Status 16 means running.
-            if instance_status == 16:
-                # Do not look for an exact match for location, because the
-                # desired location could be us-east-1, but the actual
-                # location could be 'us-east-1b'.
-                if location == "*" or instance_location.find(location) > -1:
-                    if machine_type == "*" or instance_type == machine_type:
-                        machines.append(public_dns)
+    response = ec2.describe_instances(
+        Filters=[
```
deanwetherby, line 58: Using Filters to cut down on the number of instances we have to process.

pipeline/util.py
```diff
@@ -41,36 +42,40 @@ def get_s3_buckets():
...
+                'Values': [machine_type]
+            }
+        ]
+    )
+    
+    machines = [
+        instance['PublicDnsName'] 
+        for reservation in response['Reservations']
+        for instance in reservation['Instances'] 
+        ]
```
deanwetherby, line 78: Using list comprehension to extract the machine dns information from reservation instances. I did see a note in the code about location matches so fully expecting feedback with regard to those issues.

pipeline/util.py
```diff
@@ -41,36 +42,40 @@ def get_s3_buckets():
...
-                        machines.append(public_dns)
+    response = ec2.describe_instances(
+        Filters=[
+            {
+                'Name': 'instance-state-name',
+                'Values': ['running']
+            },
+            {
+                'Name': f'tag:{tag_name}',
+                'Values': [f'*{tag_value}*']
```
deanwetherby, line 65: Doing wildcard searches on tag values, both prefix and suffix wildcards. So we can search on "CORA" and still get back "CORA 3.5" or "my-CORA-machine".

pipeline/util.py
```diff
@@ -41,36 +42,40 @@ def get_s3_buckets():
...
-#  have multiple sets of machines
-def get_aws_machines(machine_type='p2.xlarge', location='us-east-1'):
-    """ Look on AWS and determine all the machines that we have running
-    AWS that we can use. The assumption is that we are looking for machines
+def get_aws_machines(
+        machine_type: str='p2.xlarge',
+        location: str='us-east-1',
+        tag_name: str='Name',
+        tag_value: str=''
+    ) -> List[str]:
```
deanwetherby, line 49: Added type hints for function arguments and return value

pipeline/util.py
```diff
@@ -41,36 +41,40 @@ def get_s3_buckets():
     return buckets
 
 
-# TODO:  Add ability to look for particular tags (key,value) because might
-#  have multiple sets of machines
-def get_aws_machines(machine_type='p2.xlarge', location='us-east-1'):
-    """ Look on AWS and determine all the machines that we have running
-    AWS that we can use. The assumption is that we are looking for machines
+def get_aws_machines(
+        machine_type: str='p2.xlarge',
```
cedorman, line 45: Maybe we should take this opportunity to change this to instance_type, since that is terminology by EC2

pipeline/util.py
```diff
@@ -41,36 +41,40 @@ def get_s3_buckets():
...
+        location: str='us-east-1',
```
cedorman, line 46: probably should use region

pipeline/util.py
cedorman: This is taken care of by region_name = location.

cedorman (approved):

PR #11, opened 2021-04-07
Cora 3.5 submission

cedorman:
Combined opics and cora branches for changes for 3.5.


deanwetherby (approved):

configs/opics/mcs_config.ini
```diff
@@ -0,0 +1,9 @@
+[MCS]
+debug=false
+metadata=oracle
+evaluation=true
+s3_bucket=evaluation-images
+s3_folder=eval-3.5
+team=opics
+evaluation_name=eval_3-5
```
ThomasSchellenbergNextCentury, line 8: Should we have separate config files for each performer, evaluation, and metadata tier combination? If so, we should probably give them more descriptive filenames.

cora_task_runner.py
```diff
@@ -0,0 +1,151 @@
...
+# TASK_FILE_PATH = "tasks_single_task.txt"
+TASK_FILE_PATH = "tasks_juliett.txt"
+
+
+# Config file location
+config_on_local = "config_level2.ini"
+remote_config_file_directory = "/home/ubuntu/workspace/GenPRAM.jl/GenAgent/omg/"
+
+
+class CoraRunTasks:
```
ThomasSchellenbergNextCentury, line 25: The CoraRunTasks and OpicsRunTasks classes seem to have more similarities than differences. Should we make a common class for code shared between the two of them? Is that already on the roadmap?

cora_task_runner.py
```diff
@@ -0,0 +1,151 @@
...
+        self.log.info(f"Number of machines {len(self.available_machines)}")
+        self.log.info(f"Machines available:  {self.available_machines}")
+
+    def run_xstartup(self):
+        ''' Start X Server on all the machines.  Note:  Not parallelized'''
+        for machine in self.available_machines:
+            cmd = "/home/ubuntu/start_cora_docker.sh"
+            return_code = util.shell_run_background(machine, cmd, self.log)
+            print(f"Tried to start docker on CORA machine {machine}" +
+                  f" Result: {return_code}")
```
ThomasSchellenbergNextCentury, line 109: We're calling print here, and on line 129, but using `self.log` elsewhere. Not sure if that's an issue or not.

cora_task_runner.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def kill_and_restartX(self):
+        for machine in self.available_machines:
+            bs = XServerStartup(machine, self.log)
+            bs.kill_and_restart()
```
ThomasSchellenbergNextCentury, line 114: What does "bs" mean in this context? Or maybe it's just the obvious :)

cora_task_runner.py
brianpippin: We definitely want to use one for consistency sake I would think?

configs/opics/mcs_config.ini
cedorman: I thought about that, but the advantage of doing it this way is that it can be dropped into the appropriate place on the remote machine and it just works.   The different performers use different file names, so having an opics directory (with mcs_config.ini) makes sense and then having a cora directory (with config_level2.ini) makes sense. Otherwise, you have to have multiple file and then have to have a separate config saying what to rename them to.

cora_task_runner.py
cedorman: Yes, they should.  I'll add a issue.  It's hard to tell what is common and what is not while developing.  Makes sense to go back and review / clean up more.

cora_task_runner.py
cedorman: I probably had a good reason, but I have changed the names to be meaningful

cora_task_runner.py
cedorman: Changed in all the files.

PR #14, opened 2021-06-03
Mcs 676 ray actor

cedorman:
First Cut at Ray Actor.   See discussion of how it works at:  https://nextcentury.atlassian.net/wiki/spaces/MCS/pages/2156757749/BoK


rartiss55:
> First Cut at Ray Actor. See discussion of how it works at: https://nextcentury.atlassian.net/wiki/spaces/MCS/pages/2156757749/BoK

Would it make sense to reference the body of knowledge link in a comment or in the README?

autoscaler/ray_baseline_aws.yaml
```diff
@@ -0,0 +1,50 @@
+#
+#   Configuration file for:
+#         Team:   Baseline
+#
+# This file is a modification of: https://github.com/ray-project/ray/blob/master/python/ray/autoscaler/aws/example-full.yaml
+
+cluster_name: baseline
```
rartiss55, line 7: So I tried to run "ray up autoscaler/ray_baseline_aws.yaml" and it is prompting me that there is no head node found and if I'd like to launch a new cluster. I felt like since 'cache_stopped_nodes' is set to 'True' that's not what's supposed to happen -- does it have anything to do with us running from different machines?

cedorman: Yes, it appears that it keeps track of the cluster on your computer somewhere (havne't figured out where yet).   When you try to do something that deals with the cluster, like 'ray submit' or 'ray monitor', it uses some cached values such as the ip address and ports.   So, I think that the cluster is per-person.    You and I can both have the cluster up at same time;  I guess that has advantages and disadvantages.

cedorman: Also, yes, will put link to BoK in README

rartiss55: Ah okay, I added a note about how clusters work per user in the BoK.

autoscaler/ray_baseline_aws.yaml
```diff
@@ -0,0 +1,50 @@
...
+# If needed, can copy files to remote machines.  However, they should be configured to run from a script,
+# so this should not be needed
+file_mounts: {
+}
+
+setup_commands:
+  - echo 'export PATH="$HOME/anaconda3/envs/tensorflow_p36/bin:$PATH"' >> ~/.bashrc
+  # Can't seem to use the default which uses a wheel, which doesn't work on AWS Ubuntu ????
+  # - source activate pytorch_latest_p37
+  # - pip install -U ray==1.2.0
```
rartiss55, line 37: Trying this from scratch, I had to uncomment the install ray line.

pipeline_ray.py
```diff
@@ -0,0 +1,141 @@
...
+            done, not_done = ray.wait(not_done)
+            for done_ref in done:
+                result = ray.get(done_ref)
+                self.log.info(f"Result of {done_ref}:  {result}")
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description='Run scenes using ray')
+    parser.add_argument(
+        'execution_config_file',
```
rartiss55, line 120: Since this wasn't mentioned in the BoK -- do I need to specify running these two arguments? 

I'd assume for testing purposes, -mcs_config_file would be mcs_config_baseline_level2.ini. I'm assuming you were using baseline_local.ini for -execution_config_file, but you have a gravity.sh script referenced in there that I don't think is checked in -- do I need that?

autoscaler/ray_baseline_aws.yaml
```diff
@@ -0,0 +1,50 @@
+#
+#   Configuration file for:
+#         Team:   Baseline
+#
+# This file is a modification of: https://github.com/ray-project/ray/blob/master/python/ray/autoscaler/aws/example-full.yaml
+
+cluster_name: baseline
+initial_workers: 2
```
rartiss55, line 8: Does 'initial_workers' actually do anything? I would've expected 2 worker nodes on default but there are none. Although I've seen it mentioned in some copies of config files, I didn't see anything in the docs.

autoscaler/ray_baseline_aws.yaml
rartiss55: Looking at this again (since our BoK says we're using 1.3.0), did your comment here mean 1.3.0 didn't work because it needed a wheel?

autoscaler/ray_baseline_aws.yaml
cedorman: You're right.  for some reason, that's not working.   However, I think that this pull request is about the actor, rather than the yaml file.  MCS-672 should be about scaling and yaml.

pipeline_ray.py
cedorman: Good catch.  Did not add that to BoK.  Will do that.  The 'gravity.sh' I am using is a no-op.  It just prints 'got to here' and returns.  The scripts is MCS-701.

autoscaler/ray_baseline_aws.yaml
cedorman: I set it so 1.3.0, committed and pushed.

cedorman: The comment about the wheel was for the 'source activate pytorch...'.  But, that is not relevant, so removing both those lines.

configs/baseline_ray.ini
```diff
@@ -0,0 +1,13 @@
+#
+# Execution configuration for running Ray for Baseline locally
```
ThomasSchellenbergNextCentury, line 2: Maybe I missed it somewhere, but what's the difference between baseline_local and baseline_ray?

autoscaler/ray_baseline_aws.yaml
```diff
@@ -0,0 +1,48 @@
+#
+#   Configuration file for:
+#         Team:   Baseline
```
ThomasSchellenbergNextCentury, line 3: For clarification, I'm assuming any file with "baseline" in its name is for running the baselines?

ThomasSchellenbergNextCentury: And we need to define a separate autoscaler for each performer?

configs/baseline_ray.ini
ThomasSchellenbergNextCentury: I can see that the run_script and scene_location in each file are different, but I'm not sure about the intent of the files.

pipeline/tests.py
```diff
@@ -61,5 +55,33 @@ def test_get_aws_machines():
...
+
+def run_background_task_locally():
+    logger = configure_base_logging("testlog")
+    run_command = 'while :; do sleep 1; done &'
+    util.run_command_and_return(run_command, logger)
+    sleep_running = util.check_if_process_running(run_command, True)
+    if sleep_running:
+        print("sleep proc found ")
+    else:
+        print("sleep proc not found ")
```
ThomasSchellenbergNextCentury, line 76: What's the point of this function? I can't find where it's used, and it's just... sleeping, right?

pipeline/util.py
```diff
@@ -193,5 +204,33 @@ def copy_file_from_aws(machine_dns, file_name, log=None):
...
+    Check if there is any running process with a command line that matches
+    the passed process_name.  psutil.process_iter returns Process objects.
+    We look in the full command line.  Note that the commandline is a list,
+    so you have to be careful with the match.
+
+    Example:  you start a process with 'bash -s "while :; do sleep 1; done &"'
+    The command line looks like:
+           ['/bin/bash', '-c', 'while :; do sleep 1; done &']
+    So, you cannot look for the entire passed command, just parts of it.
+    '''
```
ThomasSchellenbergNextCentury, line 222: Some good comments in this file.

pipeline_ray.py
```diff
@@ -0,0 +1,146 @@
...
+        # Run the script on the machine
+        cmd = f'{self.run_script} {mcs_config_filename} {scene_config_filename}'
+        print(f"In run scene.  Running {cmd}")
+
+        # TODO:  Return more information from the script about how it went and results
+        ret = os.system(cmd)
+
+        # TODO:  Remove the mcs_config.ini file since it may have AWS information in it
+
+        # TODO:  Long term.  copy the files to S3???
```
ThomasSchellenbergNextCentury, line 54: Might be helpful to create JIRA tickets for each of these TODOs and add the corresponding ticket labels (MCS-123 or whatever) with the comments here, to track incomplete tasks more easily.

pipeline_ray.py
```diff
@@ -0,0 +1,146 @@
...
+    return parser.parse_args()
+
+
+if __name__ == '__main__':
+    args = parse_args()
+
+    # TODO:  If running local, do ray.init().  If doing remote/cluster, do (address='auto').  Add
+    # command line switch to determine which to use
+    # ray.init(address='auto')
+    ray.init()
```
ThomasSchellenbergNextCentury, line 135: Same comment as above, about creating a JIRA ticket for this TODO.

pipeline_ray.py
```diff
@@ -0,0 +1,146 @@
...
+        self.scene_files_list = []
+
+        date_str = util.get_date_in_file_format()
+        self.log = logger.configure_base_logging(date_str + ".log")
+        self.log.info("Starting run scenes")
+
+        self.get_scenes()
+        self.run_scenes()
+
+    def read_mcs_config(selfs, mcs_config_filename: str):
```
rartiss55, line 79: typo here -- self instead of selfs

pipeline_ray.py
rartiss55: Ah okay, thanks for clarifying. I tested this locally and it seems fine since we have another ticket to address actually running things.

autoscaler/ray_baseline_aws.yaml
rartiss55: Re: MCS-672, I think these BoK tasks got a bit confusing -- most of them were just about writing little scripts that didn't necessarily have to do with specifics with the pipeline yet, but more building our knowledge base on ray.io and updating the confluence page. Although the description for MCS-676 explicitly mentions the pipeline, so this one is a bit different. Anyway, I know you have MCS-701, but we likely need additional tickets to cover things like testing this existing config file/amending it if needed, tweaking the pipeline file so that we're taking full advantage of the autoscalar instead of just relying on a single worker, more thorough testing on AWS, anything else.

Leaving a reminder here to chat at standup tomorrow about this in case I forget :)

configs/baseline_ray.ini
rartiss55: Not Clark, but I interpreted this as this is the config file meant for running on the EC2 instances and the local one is for running on your own single machine. 

Nice catch though looking at this again -- comment in this file is kind of misleading and the name could be clearer. We don't even really need two checked in since everyone's local one would be different anyway, and even the clusters on EC2 may be unique to user (although there's probably some workaround for that).

pipeline_ray.py
cedorman: Yes, solutions and history files should be own issues.

configs/baseline_ray.ini
cedorman: One of the issues with Ray is that it's pretty customizable.  The idea was to provide configuration files that show how to run it in different situations / environments.   baseline_local is running on your local machine.  

baseline_ray is mis-named, and the comment is wrong.  good catch.  It should be baseline_aws since the values in there refer to locations on the AWS machine.   I will re-name and change the comment.   

In retrospect, it was a mistake to have this file in there, since this pull request is for the Actor and not for configuration.  I shouldn't have anything about AWS in this PR at all.

autoscaler/ray_baseline_aws.yaml
cedorman: Yes, things with baseline are for running the baseline.  

It's an open question about what does and does not need to be for each performer.  I **think** that this can be made generic, and so would apply to all.  The catch is what happens in the setup commands section.  If we can move everything to performer-specific execution config files, then we should not need that.

pipeline/tests.py
cedorman: The classes and functions in the pipeline/ directory are largely historical at this point.  I moved everything into there while starting the Ray part;  Ray actually replaces the vast majority of the code.  

The point of this code was to show how you can, in fact, start a process in the background on a remote machine.  It's surprisingly difficult since python seems to want to do thing synchronously and wants to kill the process that started the command.

pipeline_ray.py
cedorman: Fixed.

pipeline_ray.py
cedorman: Done.

configs/baseline_ray.ini
ThomasSchellenbergNextCentury: Cool, thanks for the explanation.

pipeline/tests.py
ThomasSchellenbergNextCentury: Thanks, that makes sense. If you want to keep this code around, would you mind adding a comment to the code with an explanation?

autoscaler/ray_baseline_aws.yaml
ThomasSchellenbergNextCentury: Sounds like a plan, and I'm sure we can figure out how to make it work!

pipeline_ray.py
ThomasSchellenbergNextCentury: I just assumed Clark had multiple personalities

pipeline/tests.py
cedorman: Moved old pipeline code to directory called old_pipeline (duh!) and added a README explaining it.

PR #16, opened 2021-06-21
Mcs 719 make actor scale

rartiss55:
PR for Clark's baseline code so that we can get this merged to main.


autoscaler/ray_baseline_aws.yaml
```diff
@@ -4,45 +4,53 @@
 #
 # This file is a modification of: https://github.com/ray-project/ray/blob/master/python/ray/autoscaler/aws/example-full.yaml
 
-cluster_name: baseline
-initial_workers: 2
-max_workers: 20                        # 2 is the default
+cluster_name: baseline-clark-3
+max_workers: 8                       # 2 is the default
```
ThomasSchellenbergNextCentury, line 8: Why the reduction in max workers?

pipeline_ray.py
```diff
@@ -71,13 +66,15 @@ def __init__(self, args):
 
         self.scene_files_list = []
 
-        date_str = time_utils.get_date_in_file_format()
-        self.log = logger.configure_base_logging(date_str + ".log")
-        self.log.info("Starting run scenes")
+        date_str = time.strftime('%Y-%m-%d', time.localtime(time.time()))
+        print(f"Starting run scenes {date_str}")
```
ThomasSchellenbergNextCentury, line 70: Why the change from using the logger to using print?

autoscaler/ray_baseline_aws.yaml
cedorman: This is the for the baseline runs, which are pretty quick.  Determining how many works to use balances a couple of different things

- There is a definite lag between starting a machine and when it appears and is useful.  
- Machines are re-used, so almost no lag when doing second run on a machine
- You are charged per hour, starting with the launch of the machine 

So if it takes 5 minutes to start and then a run takes 30 seconds to do a run (which is about what it takes for baseline), then it doesn't make sense to start a lot of machines.

pipeline_ray.py
cedorman: Ray doesn't use the logger, so you end up with a mishmash of logged outputs and stdout outputs and stderr (if there are errors), and (worse) they are not always in order.  Using print makes them all come out in the same place and in order.  Using the logger correctly is MCS-725.

ThomasSchellenbergNextCentury (approved):

PR #17, opened 2021-06-22
added baseline ray script

kdrumm:


deploy_files/baseline/ray_script.sh
```diff
@@ -0,0 +1,41 @@
...
+
+# Will be passed in two things:
+#    mcs_configfile  scene_file
+mcs_configfile=$1
+scene_file=$2
+
+echo "Running Baseline with config $mcs_configfile and scene $scene_file"
+
+
+# TODO:  Check that files exist
```
rartiss55, line 11: @cedorman do we have/need tickets for these todos?

cedorman: Added.

PR #18, opened 2021-06-23
Mcs 715 opics

kdrumm:
PR is ready now that dependent branch has been merged


aws_scripts/opics_ray_setup.sh
```diff
@@ -0,0 +1,9 @@
+#This file is not meant to executed.  It is intended to be a helper to know what commands should be run.
```
cedorman, line 1: probably should not be a .sh file if not supposed to be executed.  maybe .txt?

configs/mcs_config_opics_oracle.ini
```diff
@@ -1,8 +1,13 @@
...
+evaluation = false
+
+s3_bucket=evaluation-images
+s3_folder=eval-35-test
+
+;aws_access_key_id=
+;aws_secret_access_key=
+
+team=opics
+evaluation_name=eval375
```
cedorman, line 13: We should coordinate with Brian and come up with a standard naming scheme and document it.   The evaluation_name is combined with team, metadata level, and scene;  I think eval375 is a different scheme than previous and conflicts with the s3 folder name. I added MCS-729 .

deploy_files/opics/ray_script.sh
```diff
@@ -0,0 +1,50 @@
...
+# Pre setup - potentially move to run once per machine later
+echo Starting X Server
+#sudo nvidia-xconfig --use-display-device=None --virtual=1280x1024 --output-xconfig=/etc/X11/xorg.conf --busid=PCI:0:30:0
+sudo /usr/bin/Xorg :0 &
+echo "Sleeping for 20 seconds to wait for X server"
+sleep 20
+echo "Sleep finished"
+
+#below is modeled from baseline but I'm not sure if its needed here.
+#rm -f $SCENE_DIR/validation/*
```
cedorman, line 23: You need to review the internals of the eval.py (performer code that does the scene) to determine this.  Highly perfomer-dependent:   Baseline by default processes all the scene files in a particular directory;  CORA runs a single file that is passed in.

deploy_files/opics/ray_script.sh
```diff
@@ -0,0 +1,50 @@
...
+rm -rf $SCENE_DIR/*
+echo Moving scene_file=$scene_file to $SCENE_DIR
+cp $scene_file $SCENE_DIR/
+
+echo "Making temporary copy of config file ($mcs_configfile -> $TMP_CFG_FILE)"
+cp $mcs_configfile $TMP_CFG_FILE
+echo Removing old config file at $EVAL_DIR/mcs_config.ini
+rm $EVAL_DIR/mcs_config.ini
+echo Moving temporary config file to config location
+mv $TMP_CFG_FILE $EVAL_DIR/mcs_config.ini
```
cedorman, line 41: Why do this 2 step copy?    The $mcs_configfile is a file on the disk.  Why not copy it directly to $EVAL_DIR/mcs_config.ini

deploy_files/opics/ray_script.sh
```diff
@@ -0,0 +1,50 @@
...
+
+echo Starting Evaluation:
+echo 
+cd $EVAL_DIR && source activate mcs_opics && python3 eval.py  --scenes $SCENE_DIR
+
+# Copy the results to the right place
+echo Copying results to /tmp/results/
+mkdir -p /tmp/results/
+cp $EVAL_DIR/SCENE_HISTORY/* /tmp/results/
```
cedorman, line 50: add newline at end of file

aws_scripts/opics_ray_setup.sh
```diff
@@ -0,0 +1,9 @@
...
+#Commented commands were used in creation of the AMI, but should not be needed unless a new AMI is created.
+
+ray up -y autoscaler/ray_opics_aws.yaml
+ray rsync_up -v autoscaler/ray_opics_aws.yaml configs/ '~/configs/'
+#ray rsync_up -v autoscaler/ray_opics_aws.yaml deploy_files/opics/ '~'
+#ray rsync_up -v autoscaler/ray_opics_aws.yaml pipeline '~'
+#ray rsync_up -v autoscaler/ray_opics_aws.yaml scenes_single_scene.txt '~'
+ray submit autoscaler/ray_opics_aws.yaml pipeline_ray.py configs/opics_aws.ini configs/mcs_config_opics_oracle.ini
```
cedorman, line 9: add newline at end of file

deploy_files/opics/ray_script.sh
kdrumm: This keeps the original file around, but if the original file is the same as our destination, we don't lose the file when we delete the "old" file.  This may be unnecessary with ray, but was useful testing.

deploy_files/opics/ray_script.sh
kdrumm: done

aws_scripts/opics_ray_setup.sh
kdrumm: done

deploy_files/opics/ray_script.sh
kdrumm: I removed it.  eval.py takes a directory as a parameter which we clear earlier in the script.

aws_scripts/opics_ray_setup.sh
kdrumm: done

rartiss55 (commented):
LGTM -- only thing that jumped out at me is the comment at the top of the ray_opics_aws.yaml (should say 'Team: OPICS ' instead of 'baseline').

Thanks for the slack comments btw for things to discuss -- I myself am struggling with when to call these tickets 'done' because of some of the points you listed.

autoscaler/ray_opics_aws.yaml
```diff
@@ -0,0 +1,56 @@
+#
+#   Configuration file for:
+#         Team:   OPICS
+#
+# This file is a modification of: https://github.com/ray-project/ray/blob/master/python/ray/autoscaler/aws/example-full.yaml
```
ThomasSchellenbergNextCentury, line 5: There's a lot of configuration in here that's shared across the performer-specific files. It might be nice if, as a future (lower priority) ticket, we could instead:
- Have a single YAML file containing the common configuration properties
- Have a YAML file containing the configuration properties specific to each performer
- Have a script that, when run, combines the common config file with each performer-specific config file and generates the final files (the script should be pretty simple to make)
- Run the script as a pre-commit action whenever the config files change

We might be able to do something similar with a few of the other performer-specific files, like the ini files. Thoughts?

ThomasSchellenbergNextCentury (approved):

autoscaler/ray_opics_aws.yaml
kdrumm: I do like the idea of not having to duplicate a bunch of configuration.  I think we should investigate this further.  I'm not exactly sold on make it a pre-commit action.  My thought would be to make that part of the execution workflow itself.  Its worth discussing at a later time though.

PR #19, opened 2021-06-25
MCS-714 run script for mess

rartiss55:
Putting this in as draft -- wanted to tweak the AMI a bit and test some more.

Additionally, this only seems to work after the first run for newly created machines (where it times out -- I think Clark mentioned something about that, so that might be normal?)

### DIFF
--- a/autoscaler/ray_mess_aws.yaml
+++ b/autoscaler/ray_mess_aws.yaml
@@ -4,8 +4,10 @@
 #
 # This file is a modification of: https://github.com/ray-project/ray/blob/master/python/ray/autoscaler/aws/example-full.yaml
 
-cluster_name: mess-375test-ra
+cluster_name: mess-375-test
 max_workers: 8                       # 2 is the default
+
+idle_timeout_minutes: 10
 
 # Note that ray only seems to use the first availability zone, so if there are issues with that 
 # one, just need to put a different one first.
@@ -23,14 +25,14 @@ available_node_types:
     resources: {"CPU": 4, "GPU": 1}
     node_config:
       InstanceType: p2.xlarge
-      ImageId: ami-0c81f616d7ca4e0e3 # beta
+      ImageId: ami-0aabadbc0dd018b89 # AMI built from 'mess-35-beta'
   ray.worker.default:
     min_workers: 0
     max_workers: 3
     resources: {"CPU": 4, "GPU": 1}
     node_config:
       InstanceType: p2.xlarge
-      ImageId: ami-0c81f616d7ca4e0e3 # beta
+      ImageId: ami-0aabadbc0dd018b89 # AMI built from 'mess-35-beta'
 
 auth:
   ssh_user: ubuntu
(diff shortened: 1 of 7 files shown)


kdrumm (commented):
I know this is just a draft, but it looks good to me so far.

autoscaler/ray_mess_aws.yaml
```diff
@@ -0,0 +1,58 @@
...
+max_workers: 8                       # 2 is the default
+
+idle_timeout_minutes: 10
+
+# Note that ray only seems to use the first availability zone, so if there are issues with that 
+# one, just need to put a different one first.
+provider:
+  type: aws
+  region: us-east-1
+  availability_zone: us-east-1b, us-east-1c, us-east-1d, us-east-1a
```
kdrumm, line 17: This doesn't work for me unless I remove the spaces between entries.

rartiss55: Odd -- well I removed them, hopefully that fixes things.

autoscaler/ray_mess_aws.yaml
```diff
@@ -0,0 +1,62 @@
+#
+#   Configuration file for:
+#         Team:   MESS
+#
+# This file is a modification of: https://github.com/ray-project/ray/blob/master/python/ray/autoscaler/aws/example-full.yaml
+
+cluster_name: mess-375-test
+max_workers: 8                       # 2 is the default
+
+idle_timeout_minutes: 10
```
rartiss55, line 10: I added a timeout since I needed extra time to log onto the worker nodes and troubleshoot some issues --decided to leave it in.

configs/mcs_config_mess_level2.ini
```diff
@@ -0,0 +1,14 @@
...
+;aws_access_key_id=
+;aws_secret_access_key=
+
+; update evaluation property and s3 folder for running eval
+evaluation=false
+s3_bucket=evaluation-images
+s3_folder=eval-35-test
+
+team=mess
+evaluation_name=eval_3-75
```
rartiss55, line 14: I'm pretty sure this is the evaluation naming pattern we need for everything to work correctly, but we can follow up with MCS-729.

autoscaler/ray_mess_aws.yaml
```diff
@@ -0,0 +1,62 @@
...
+
+# If needed, can copy files to remote machines. Commenting out for now, since
+# a copy is of this script is included in the AMI. 
+file_mounts: {
+  #"/home/ubuntu/ray_script.sh": "./tmp_pipeline_ray/ray_script.sh"
+}
+
+setup_commands:
+  - pip install -U ray==1.3.0
+  - rm -rf ~/.aws/
```
rartiss55, line 52: I don't have the credentials file on my AMI, but still had this directory, so added this in for consistency since it will likely be in other ray config files anyway.

ThomasSchellenbergNextCentury (approved):
Didn't run it, but the code LGTM.

PR #20, opened 2021-06-25
Mcs 713 Cora

cedorman:
Implementation of the CORA scaling using Ray.


kdrumm (commented):
LGTM, although now that some other PRs were merged, we should merge dev in.  Github seems to not be factoring that into the diff.

autoscaler/ray_cora_aws.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+cluster_name: cora-clark-v4
+max_workers: 8                       # 2 is the default
+
+# Note that ray only seems to use the first availability zone, so if there are issues with that 
+# one, just need to put a different one first.
+provider:
+  type: aws
+  region: us-east-1
+  availability_zone: us-east-1c, us-east-1b, us-east-1d, us-east-1a
```
kdrumm, line 15: This doesn't work for me unless we remove the spaces.

ThomasSchellenbergNextCentury (approved):
Didn't run it, but the code LGTM. Good use of comments.

autoscaler/ray_cora_aws.yaml
cedorman: Thanks.  Noticed that was fixed in your alignment.

autoscaler/ray_cora_aws.yaml
```diff
@@ -0,0 +1,56 @@
+#
+#   Configuration file for:
+#         Team:   Baseline
```
rartiss55, line 3: Comment update -- should say "CORA" here.

PR #22, opened 2021-06-28
Common start script

kdrumm:
There are some differences between our scripts that we may need to resolve before this will work.  The one i know in particular is the starting location on the head node for the scene files.  This script assumes ~/scenes/tmp/.

Additionally, @rartiss55  brought up that we may want to be careful where we put the pipeline code.  Currently its upload directly to the home directory.  In the end, individual ray_script.sh files can move the files to wherever they are needed once we establish where they should start.

All code that is eval specific on the run_eval.sh script is moved to the end so if we want a more generic pipeline script, it'll be easier to move.


cedorman (commented):
Is this supposed to coordinate between all 4 (baseline, MESS, OPICS, CORA)?  My CORA review is still on-going, and should be aligned with this.

aws_scripts/run_eval.sh
```diff
@@ -0,0 +1,76 @@
...
+rm -rf $TMP_DIR/*
+
+ORIG_PWD=$PWD
+
+
+#cp -R configs $TMP_DIR
+cp -R deploy_files/${MODULE}/* $TMP_DIR/
+#mkdir -p $TMP_DIR/scenes/tmp
+#cp -R $LOCAL_SCENE_DIR/* $TMP_DIR/scenes/tmp/
+RAY_CONFIG="autoscaler/ray_${MODULE}_aws.yaml"
```
ThomasSchellenbergNextCentury, line 25: Are these commented-out commands still used sometimes? If not, do you want to delete them?

autoscaler/ray_opics_aws.yaml
```diff
@@ -42,7 +42,9 @@ head_node_type: ray.head.default
 # If needed, can copy files to remote machines.  However, they should be configured to run from a script,
 # so this should not be needed
 file_mounts: {
+  "/home/ubuntu/ray_script.sh": ".tmp_pipeline_ray/ray_script.sh"
 }
+# file_mounts_sync_continuously: True
```
ThomasSchellenbergNextCentury, line 47: Is this still needed?

kdrumm:
> Is this supposed to coordinate between all 4 (baseline, MESS, OPICS, CORA)? My CORA review is still on-going, and should be aligned with this.

That's the idea.  Anyone should easily be able to run an eval by running this script with the name of the team and a path to directory with scene files.  It will also ensure the current version of the code will be pushed.  I think the only issue to reconcile is where the scenes are expected to be pushed on the head node.  This is should be extremely easy to reconcile.  (I believe some are ~/scenes/tmp and some are ~/scenes)

It does push the pipeline code to ~.  @rartiss55 brought up yesterday that we may want to be more careful where we put that.  Then the ray_script.sh can move it to wherever it needs to be.

autoscaler/ray_opics_aws.yaml
kdrumm: I'd love to say they are just there to test you, but really I just get used to not reading comments like this in my code.  Thanks for catching them.  I removed them.

cedorman:
> This is should be extremely easy to reconcile. (I believe some are ~/scenes/tmp and some are ~/scenes)

I can modify my CORA code later today or tomorrow morning and test.

kdrumm:
> > This is should be extremely easy to reconcile. (I believe some are ~/scenes/tmp and some are ~/scenes)
> 
> I can modify my CORA code later today or tomorrow morning and test.

If it works as is (without the common script), feel free to leave the reconciliation to me.  There is also AWS security that needs reconciled.

autoscaler/ray_opics_aws.yaml
ThomasSchellenbergNextCentury: Thanks for making me feel useful! :)

brianpippin:
I did a review and nothing stuck out to me so far with this.

aws_scripts/run_eval.sh
```diff
@@ -0,0 +1,71 @@
...
+cp -R deploy_files/${MODULE}/* $TMP_DIR/
+
+RAY_CONFIG="autoscaler/ray_${MODULE}_aws.yaml"
+
+ray up -y $RAY_CONFIG
+wait
+
+# We should copy all the pipeline code, but at least opics needs it in a special folder.  Should ray_script handle that?  
+# Should we run
+ray rsync_up -v $RAY_CONFIG pipeline '~'
```
rartiss55, line 27: When everything is merged, do we want to make this "git pull" instead or something like that? I believe the code is already checked out on most if not all the AMIs -- wondering if that makes more sense.

kdrumm: I would argue no.  That would lock us in to a specific version making any bug fixes or enhancements hard.  If we ever wanted to run different versions this helps.  The user already needs mcs-pipeline checked out on their own machine anyway.

autoscaler/ray_opics_aws.yaml
```diff
@@ -42,6 +42,7 @@ head_node_type: ray.head.default
 # If needed, can copy files to remote machines.  However, they should be configured to run from a script,
 # so this should not be needed
 file_mounts: {
+  "/home/ubuntu/ray_script.sh": ".tmp_pipeline_ray/ray_script.sh"
```
kdrumm, line 45: Just realized with this, the config pretty much assumes it is being run from the script.  That wasn't exactly intentional, although i still think running from the script is much easier as it keeps your local environment in sync with the remote environments.  

The question is, do we want this in all our configs and force everyone to use the script?

kdrumm: On further review, the idea of moving the ray_script here is not needed.  It was a forgotten result of a bunch of tests to get the script into the worker nodes.  I propose the following:

Change the idea of "deploy_files" and rename to be files that will be handled by the config.  Allow each config to file_mount their own script and possible any other needed files.

Remove "deploy_files" with the idea that we could bring it back if we ever have a bunch of files that need to be moved to just the head node.

kdrumm: Arguably, the configs and ray_script could be in the same place and moved via file_mounts (even though configs don't need to be on workers)

kdrumm:
I was unable to get cora to work, but mess, baseline and opics work now with script.

aws_scripts/run_eval.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+# This file is intended to be run on local development machine.
+# This will perform the following actions:
+#  1. Start a ray cluster for the appropriate module
+#  2. Upload the appropriate scenes
+#  3. Upload the config files
+#  4. Start the eval
+
+MODULE=$1
+TMP_DIR=.tmp_pipeline_ray
```
rartiss55, line 10: Do we still need this? IIRC this was where the file_mount path was before everything was merged.

kdrumm: This is where the generated scene list goes

rartiss55 (approved):

PR #23, opened 2021-07-01
Mcs 709 retry on results

kdrumm:
This is a much less confident PR.  First, I'd like some comments on the approach and to see if anything glaring is missing.  

My goals were: 

- To be able to get, retain (for the life of the execution) and print more status information.
- Handle and retry on the MCS timeout error
- Provide a mechanism to facilitate retries on other errors.

Currently all errors will result in up to 3 retries.  

Note: Baselines failure to move the heatmap file does not cause a return code != 0


pipeline_ray.py
```diff
@@ -95,20 +142,56 @@ def get_scenes(self):
...
+        status = RunStatus(result, output, "Success", False)
+        # Result is really the result of the copy result files (or last command in ray_script.sh)
+        if (result is not 0):
+            status.retry |= False
+            status.status = "Error"
+        if ("Exception in create_controller() Time out!" in output):
+            print(f"Timeout occured for file={scene_file_path}")
+            status.retry |= True
+            status.status = "Error: Timeout"
+        # Add more conditions to retry here
```
cedorman, line 193: How does this do with the HeatMap missing issue?   Can we identify those?

pipeline_ray.py
```diff
@@ -95,20 +142,56 @@ def get_scenes(self):
...
+        with open(str(scene_ref)) as scene_file:
+            job_id = run_scene.remote(self.exec_config['MCS']['run_script'],self.mcs_config, json.load(scene_file))
+            print(f"Retrying {scene_ref} with job_id={job_id}")
+            self.scene_statuses[job_id] = scene_status
+            not_done.append(job_id)
+
+    def get_run_status(self, result:int, output:str, scene_file_path:str) -> RunStatus:
+        status = RunStatus(result, output, "Success", False)
+        # Result is really the result of the copy result files (or last command in ray_script.sh)
+        if (result is not 0):
```
cedorman, line 186: Minor:  Don't need the parentheses

pipeline_ray.py
```diff
@@ -95,20 +142,56 @@ def get_scenes(self):
...
+        # This should probably be configurable and may need to be different depending on what errors we are detecting.
+        # This should work for a first step though.
+        num_retries = 3
         print(f"Running {len(self.scene_files_list)} scenes")
         job_ids = []
         for scene_ref in self.scene_files_list:
             with open(str(scene_ref)) as scene_file:
-                job_ids.append(run_scene.remote(self.exec_config['MCS']['run_script'],self.mcs_config, json.load(scene_file)))
+                job_id = run_scene.remote(self.exec_config['MCS']['run_script'], self.mcs_config, json.load(scene_file))
+                self.scene_statuses[job_id] = self.scene_statuses.get(job_id, SceneStatus(scene_ref, 0, "pending"))
```
cedorman, line 153: The logic is unclear here.  Under what conditions would this not go to default?  This is the first time through, so get(job_id) would never work.

pipeline_ray.py
```diff
@@ -41,20 +44,40 @@ def run_scene(run_script, mcs_config, scene_config):
...
+@dataclass
+class SceneStatus:
+    scene_file: str
+    retries: int = 0
+    status: str = ""
+    run_statuses: List[RunStatus] = field(default_factory=list)
+    
 class SceneRunner:
 
+    scene_statuses={}
```
cedorman, line 79: Should have a comment to say what this is an associative array of, i.e. that this maps Ray job_id to SceneStatus.   That's a little weird though, because on re-tries there will be a new job_id for the same SceneStatus, or rather you will have an old job_id with the original SceneStatus and a new job_id with the new status for the same scene.  

I think that it would make more sense to have scene_status map from scene file to SceneStatus.   You would have to store the job_id for the current running job in the SceneStatus.  When a job is finished and you get the done_ref, you would then have to look up for the scenestatus with that job_id.

pipeline_ray.py
```diff
@@ -95,20 +142,56 @@ def get_scenes(self):
...
+                self.scene_statuses[job_id] = self.scene_statuses.get(job_id, SceneStatus(scene_ref, 0, "pending"))
+                job_ids.append(job_id)
 
         not_done = job_ids
         while not_done:
             done, not_done = ray.wait(not_done)
             for done_ref in done:
-                result = ray.get(done_ref)
+                result,output = ray.get(done_ref)
+                run_status = self.get_run_status(result, output, scene_ref)
```
cedorman, line 161: Where is scene_ref getting set here?   I think that this is just using the scene_ref from the end of the previous loop, since it doesn't get updated.  It looks like get_run_status doens't ever use this parameter.

I think the logic should be:
     get scene_status based on done_ref
     get scene_ref from scene_status 
     get run_status from get_run_status
    update run_statuses with run_status

pipeline_ray.py
kdrumm: This does not detect heatmap.  We can certainly identify the string, but it doesn't result in an error code so I've ignored it.

pipeline_ray.py
kdrumm: For any scene, there is only one scene status.  A scene status can become the value for more than one key in the case of retries, but that is intentional.  I did it this way to avoid multiple levels of map lookups or searching a whole list.  I'm not sure I'm understanding the problem here.

pipeline_ray.py
kdrumm: scene ref was only used for printing debug output.  I updated it to come from the scene_status.

pipeline_ray.py
kdrumm: You're right.  Fixed it.

pipeline_ray.py
kdrumm: Fixed.

cedorman (approved):

PR #24, opened 2021-07-02
Mcs 752 cora rework

cedorman:
This branch is a minor tweak to the script and pointing to an AMI that works.


kdrumm:
I think this is good.  I am rerunning now (to recover my long weekend memory loss). Later I want to investigate how the missing heatmap error is different than baseline.  However, that heatmap error is probably something to be handle outside this PR.

cedorman:
Yes heatmap should be MCS-730, but requires MCS code change.  For now, we just ignore it.

kdrumm (approved):

PR #27, opened 2021-07-07
Mcs 726 727 checks in scripts

cedorman:
Added common code scripts, copying of these scripts to head nodes.   Made the ray_script.sh files as similar as possible.


kdrumm (commented):
Looks good so far.  I'm going to run a couple tests to verify.  I appreciate the cleanup and common scripts!

rartiss55:
Looks good -- tested with a couple performer configs. The only thing I noticed is that the MESS ray script had the permissions updated by mistake.

autoscaler/ray_baseline_aws.yaml
```diff
@@ -39,15 +39,17 @@ auth:
...
-  #push ray_script.sh to ALL nodes
-  "/home/ubuntu/ray_script.sh": "deploy_files/baseline/ray_script.sh"
+  "/home/ubuntu/ray_script.sh": "deploy_files/baseline/ray_script.sh",
+  "/home/ubuntu/check_passed_variables.sh": "deploy_files/check_passed_variables.sh",
+  "/home/ubuntu/start_x_server.sh": "deploy_files/start_x_server.sh"
 }
 
 setup_commands:
   - pip install -U ray==1.3.0
+  - chmod +x /home/ubuntu/ray_script.sh /home/ubuntu/check_passed_variables.sh /home/ubuntu/start_x_server.sh
```
kdrumm, line 51: Speaking of permissions from @rartiss55 's comment, shouldn't we just check thses scripts in with the right permissions and not worry about it here?  Or is that somewhat of an OS issue.

kdrumm (approved):

deploy_files/baseline/ray_script.sh
```diff
@@ -1,41 +1,39 @@
...
-echo "Running Baseline with config $mcs_configfile and scene $scene_file"
-
-
-# TODO:  Check that files exist
+# Check passed mcs_config and scene file
+source /home/ubuntu/check_passed_variables.sh
 
+EVAL_DIR=/home/ubuntu/
+SCENE_DIR="$EVAL_DIR/scenes/validation/"
+TMP_CFG_FILE="$EVAL_DIR/msc_cfg.ini.tmp"
```
ThomasSchellenbergNextCentury, line 8: Is there a specific reason why some variables are lowercase and others are uppercase? (I'm not familiar with any bash style standards.)

deploy_files/baseline/ray_script.sh
```diff
@@ -1,41 +1,39 @@
...
-
-# Run the model
+echo Clearing History at $EVAL_DIR/SCENE_HISTORY/
+rm -f $EVAL_DIR/SCENE_HISTORY/*
+echo Clearing $SCENE_DIR
+rm -rf $SCENE_DIR/*
+
+# Move files to appropriate locations
+echo Making SCENE_DIR=$SCENE_DIR
+mkdir -p $SCENE_DIR
```
ThomasSchellenbergNextCentury, line 23: Thanks for adding the `mkdir` calls!

deploy_files/baseline/ray_script.sh
cedorman: There is often a distinction between environment, constants, and shell variables (which are all uppercase) and loop or temporary variables (which are lowercase with underscores) but it's not a strong distinction.   In this case, the template I was going with had all caps for these important variables which were used throughout the script, so I thought that it was a good idea.  It is obvious that you are using these when you do and where they came from.

PR #28, opened 2021-07-07
MCS-729 standardizing output files

rartiss55:
We wanted to document and enforce the correct naming of output files -- ended up going with simply validating the config file. There is an optional flag to turn off validation if we're only testing.

Also created a follow on ticket to update some of this since we're changing the s3 output folder structure in Eval 4+.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,12 +62,12 @@ In order to test the pipeline and evaluations, the following is helpful:
 
 To run an eval, run the following command:
 ```
-aws_scripts/run_eval MODULE path/to/scene/directory [metadata_level]
+aws_scripts/run_eval MODULE path/to/scene/directory --metadata [metadata_level]
 ```
 
 There is an optional flag to disable config file validation checks if you are just running tests:
 ```
-aws_scripts/run_eval MODULE path/to/scene/directory [metadata_level] --disable_validation
+aws_scripts/run_eval MODULE path/to/scene/directory --metadata [metadata_level] --disable_validation
 ```
 
 Here is an example:
--- a/README.md
+++ b/README.md
@@ -30,6 +30,20 @@ $ source venv/bin/activate
 
 ### Run Eval Script
 
+### MCS Config File 
+
+When running an eval, there are checks to ensure that the MCS config file used is using the correct naming conventions, so that all the ingest
+and UI related functionality will work correctly (these can be turned off for testing):
+
+- **metadata** - has to be either `level1`, `level2`, or `oracle`
+- **evaluation_name** - has to be one of the following, in the exact format: `eval_3-75`, `eval_4`, `eval_5`, `eval_6`, `eval_7`, `eval_8`
+- **evaluation** - must be set to `true`
+- **team** - has to be either `mess`, `mit`, `opics`, or `baseline`
+- **s3_bucket** - should be `evaluation-images`
+- **s3_folder** - has to be the folder we store output for the current eval (right now, `eval-3.75`)
+
+If anything above changes, we will need to make sure those changes are incorporated into the ingest process/UI as needed. 
+
 #### Eval test configuration
 
 In order to test the pipeline and evaluations, the following is helpful:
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -50,6 +65,11 @@ To run an eval, run the following command:
 aws_scripts/run_eval MODULE path/to/scene/directory [metadata_level]
 ```
 
+There is an optional flag to disable config file validation checks if you are just running tests:
+```
+aws_scripts/run_eval MODULE path/to/scene/directory [metadata_level] --disable_validation
```
kdrumm, line 70: Currently metadata_level is an optional parameter.  This change makes it no longer an optional parameter.  I think we're going to need more advanced parameter handling

pipeline_ray.py
```diff
@@ -127,6 +136,36 @@ def read_mcs_config(self, mcs_config_filename: str):
...
+
+            if(mcs_config_parse['MCS']['evaluation'] != 'true'):
+                print('Error: Evaluation property in MCS config file is not set to true.')
+                valid = False
+            if(mcs_config_parse['MCS']['s3_bucket'] != self.CURRENT_EVAL_BUCKET):
+                print('Error: MCS Config file does not have the correct s3 bucket specified.')
+                valid = False
+            if(mcs_config_parse['MCS']['s3_folder'] != self.CURRENT_EVAL_FOLDER):
+                print('Error: MCS Config file does not have the correct s3 folder specified.')
+                valid = False
```
kdrumm, line 154: Why do we have config if there is only one valid option?  Should we use regex instead?  Or use the config the override the default and just remove the entry from all our checked in configs.

README.md
rartiss55: Good point -- I added some logic for the optional stuff + updated the README -- now we'll have to do something like:
run_eval.sh baseline some/path/ --metadata oracle --disable_validation

pipeline_ray.py
rartiss55: In hindsight that might make more sense for the s3_bucket to default to 'evaluation_images' in the MCS python package, but I think we'll need something more flexible for the s3_folder (I added MCS-754 for that one).

aws_scripts/run_eval.sh
```diff
@@ -36,8 +36,22 @@ DEFAULT_METADATA="level2"
...
+# Parse optional --parameters
+while [ $# -gt 0 ]; do
+    if [[ $1 == *"--"* ]]; then
+        if [ $1 == "--disable_validation" ] ; then
+            VALIDATE_CONFIG="$1"
+        else
+            uppercase_param=$(echo "$1" | tr '[:lower:]' '[:upper:]')
+            param="${uppercase_param/--/}"
+            declare $param="$2"
+        fi
```
kdrumm, line 51: I barely understand how or why this works, but this is cool magic.  It'll work for any param with a value!

kdrumm (approved):
LGTM although it might be nice to get another opinion.

pipeline_ray.py
rartiss55: A little while ago when I was updating a couple properties in our config files, I came back to this question...
at some point we probably could clean up the config directory a bit -- not sure if we want eval vs test versions checked in (I left our test s3 directory in for now), or if perhaps we want to clear this out and build a config from scratch using parameters?

Anyway, we can probably discuss some of this tomorrow, and I imagine we'll discuss the 3.5 rerun (which I didn't account for in this PR, since I'm not entirely sure what we're naming things for that).

README.md
```diff
@@ -42,12 +56,18 @@ In order to test the pipeline and evaluations, the following is helpful:
...
   * Setting the s3_folder in the MCS config file to have a suffix of -test is a good idea.  I.E. s3_folder=eval-35-test 
   * The S3 file names are generated partially by the `team` and `evaluation_name` properties in the MCS config file.  Prefixing `evaluation_name` with your initials or a personal ID can make it easier to find your files in S3.  I.E evaluation_name=kdrumm-eval375
+  * Make sure MCS config file validation is off if for testing (see commands below).
 
 #### Commands
 
 To run an eval, run the following command on your local development machine (driver):
 ```
-aws_scripts/run_eval MODULE path/to/scene/directory [metadata_level]
+aws_scripts/run_eval MODULE path/to/scene/directory --metadata [metadata_level]
```
cedorman, line 65: What's MODULE?   Should it be team name?  
Also, I think it is 'run_eval.sh' rather then run_eval

aws_scripts/run_eval.sh
cedorman: declare is a new command to me.  I still don't understand it entirely, but this helped:  https://superuser.com/questions/821094/what-is-the-difference-between-set-env-declare-and-export-when-setting-a-varia

deploy_files/mess/ray_script.sh
```diff
@@ -10,7 +10,7 @@ echo "Running MESS with config $mcs_configfile and scene $scene_file"
 # Start X
 # TODO: Check to see if the xserver is already running and do not restart (MCS-727)
 cd mcs-pipeline/xserver
-sudo python3 run_startx.py &
+sudo nohup python3 run_startx.py 1>startx-out.txt 2>startx-err.txt &
```
cedorman, line 13: This is going to have a conflict with the recent MCS-726 merge;  use that one.

cedorman (approved):

README.md
kdrumm: Module is an attempt to move towards using the pipeline for more than just evals.  Of course while the script is named 'run_eval.sh' that is less obvious.  With Thomas working on using the pipeline to create videos, hopefully the path to be more generic will become more apparent soon.

PR #30, opened 2021-07-08
Mcs 725 logging

kdrumm:
Logging updates for pipeline.  

This does not capture the head node logs.  We maybe should add that, but can do it separately.

Also note: the logs are ALWAYS pushed to S3.  Not just for evaluations.  That seems reasonable to me, but maybe not?

### DIFF
--- a/pipeline_ray.py
+++ b/pipeline_ray.py
@@ -234,12 +234,6 @@ class SceneRunner:
         logging.info(f"Scenes {self.scene_files_list}")
 
     def run_scenes(self):
-        # This should probably be configurable and may need to be different depending on what errors we are detecting.
-        # This should work for a first step though.
-        log_dir = pathlib.Path("/tmp/results/logs")
-        shutil.rmtree(log_dir, ignore_errors=True)
-        log_dir.mkdir(parents=True, exist_ok=True)
-
         num_retries = 3
         logging.info(f"Running {len(self.scene_files_list)} scenes")
         job_ids = []
--- a/pipeline_ray.py
+++ b/pipeline_ray.py
@@ -249,8 +249,8 @@ class SceneRunner:
                 logging.info(f"{len(not_done)}  Result of {done_ref}:  {result}")
 
     def write_log(self, output, scene_status:SceneStatus, log_dir:pathlib.Path):
-        #Retries is +2 to get test number because +1 for next run increment and +1 for changing to 0 base to 1 base index
-        log_file = f"{pathlib.Path(scene_status.scene_file).name}-{scene_status.retries + 2}.log"
+        
+        log_file = f"{pathlib.Path(scene_status.scene_file).name}-{scene_status.retries + 1}.log"
         log_file = log_dir.joinpath(log_file)
         logging.info(f"Writing to {log_file.absolute()}")
         with open(log_file.absolute(), "a") as f:
@@ -260,7 +260,8 @@ class SceneRunner:
     def do_retry(self, not_done, scene_status:SceneStatus):
         scene_ref=scene_status.scene_file
         with open(str(scene_ref)) as scene_file:
-            job_id = run_scene.remote(self.exec_config['MCS']['run_script'],self.mcs_config, json.load(scene_file), scene_ref.name, scene_status.retries + 1)
+            #Retries is +2 to get test number because +1 for next run increment and +1 for changing to 0 base to 1 base index
+            job_id = run_scene.remote(self.exec_config['MCS']['run_script'],self.mcs_config, json.load(scene_file), scene_ref.name, scene_status.retries + 2)
             logging.info(f"Retrying {scene_ref} with job_id={job_id}")
             self.scene_statuses[job_id] = scene_status
             not_done.append(job_id)
--- a/pipeline_ray.py
+++ b/pipeline_ray.py
@@ -22,6 +22,7 @@ from logging import config
 import ray
 import boto3
 
+# This is logging for head node during setup and distribution.  
 logging.basicConfig(level=logging.DEBUG,format="%(message)s")
 
 def push_to_s3(source_file: pathlib, bucket: str, s3_filename: str, mimetype:str='text/plain', client=None):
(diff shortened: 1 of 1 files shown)


pipeline_ray.py
```diff
@@ -7,47 +7,80 @@
...
+            s3_filename,
+            ExtraArgs={
+                'ACL': 'public-read',
+                'ContentType': mimetype,
+            }
+    )
 
 @ray.remote(num_gpus=1)
-def run_scene(run_script, mcs_config, scene_config):
+def run_scene(run_script, mcs_config, scene_config, scene_try):
```
kdrumm, line 44: I was using the try # before for filenames, but have changed most of it to timestamp.  Do we want to use this at all or should I completely remove it.  Note, its still in the local log file name but that file is cleaned up when execution is finished.

pipeline_ray.py
```diff
@@ -7,47 +7,79 @@
 #
 import argparse
 import configparser
+import datetime
 import json
+from os import mkdir
```
deanwetherby, line 12: Are you using this or just the pathlib.mkdir?

pipeline_ray.py
```diff
@@ -57,8 +89,67 @@ def run_scene(run_script, mcs_config, scene_config):
...
+    s3_filename = folder + "/" + '_'.join(
+            [eval_name, metadata,
+             team,
+             scene_name, timestamp,
+             "log"]) + '.txt'
+
+    # Might need to find way to flush logs and/or stop logging.
+    push_to_s3(log_file, bucket, s3_filename)
+
+    logging.shutdown()
```
deanwetherby, line 107: I hadn't used logging shutdown before. Interesting to know that this exists.

pipeline_ray.py
kdrumm: whoops.. Not using this anymore.

pipeline_ray.py
```diff
@@ -7,47 +7,79 @@
...
+                'ContentType': mimetype,
+            }
+    )
 
 @ray.remote(num_gpus=1)
-def run_scene(run_script, mcs_config, scene_config):
+def run_scene(run_script, mcs_config, scene_config, scene_try):
     """ Ray """
+    scene_name=scene_config.get("name","")
+    timestamp=datetime.datetime.now().strftime("%Y%m%d-%H%M%S")
```
ThomasSchellenbergNextCentury, line 45: No spacing around `=`. Same on line 95 (assigning the `eval_name`)

pipeline_ray.py
```diff
@@ -7,47 +7,79 @@
...
+    config = configparser.ConfigParser()
+    config.read_string("\n".join(mcs_config))
```
ThomasSchellenbergNextCentury, line 47: Good to know about the configparser module!

pipeline_ray.py
```diff
@@ -7,47 +7,79 @@
...
+
+    log_dir = pathlib.Path("/tmp/results/logs")
+    log_dir.mkdir(parents=True, exist_ok=True)
+    # Do we want the timestamp?  If so, we will leave a bunch of log files on a single machien.  
+    # Do we clean up files?  If we don't do the timestamp, we will need to cleanup logs here so we 
+    # don't end up with old executes logs at the top of a new execution.  This may not really happen 
+    # in real evals, but happens when testing often.
```
ThomasSchellenbergNextCentury, line 54: IMO using the timestamps is fine

pipeline_ray.py
```diff
@@ -7,47 +7,79 @@
...
 
 import ray
-
+import boto3
+
+logging.basicConfig(level=logging.DEBUG,format="%(message)s")
+
+def push_to_s3(source_file: pathlib, bucket: str, s3_filename: str, mimetype:str='text/plain', client=None):
+    if client is None:
+        client = boto3.client('s3')
```
ThomasSchellenbergNextCentury, line 29: Rather than having the `client` as a function argument, should it be a global variable? That way it only ever needs to be instantiated once.

pipeline_ray.py
```diff
@@ -57,8 +88,67 @@ def run_scene(run_script, mcs_config, scene_config):
 
     # TODO  MCS-674:  Move Evaluation Code out of Python API (and into the pipeline)
 
+    # This seems a little dirty, but its mostly copied from MCS project.
+    bucket = config.get("MCS", "s3_bucket")
```
rartiss55, line 92: I ended up using configparser to validate the mcs config file -- wondering if we should use something like that here as well and just pass that object along instead of the List(string) representation of mcs_config (sorry I didn't think of this before).

pipeline_ray.py
kdrumm: The goal here is initialize it if needed with flexibility for later when we use it more than once per worker.  If we're going to use it more than once, we should make it in the run_scene function because this is running on a ray worker.

pipeline_ray.py
kdrumm: My thought would be merge them as is and then add a ticket to reconcile and rethink it if we're repeating tasks/code due to parallel tickets.  I think it was original passed as a list of strings so we could recreate the file remotely which still needs to be done.

pipeline_ray.py
```diff
@@ -7,58 +7,149 @@
...
     """ Ray """
+    scene_name = scene_config.get("name","")
+    timestamp = datetime.datetime.now().strftime("%Y%m%d-%H%M%S")
+    config = configparser.ConfigParser()
+    config.read_string("\n".join(mcs_config))
+
+    log_dir = pathlib.Path("/tmp/results/logs")
+    log_dir.mkdir(parents=True, exist_ok=True)
+    # Do we want the timestamp?  If so, we will leave a bunch of log files on a single machien.  
+    # Do we clean up files?  If we don't do the timestamp, we will need to cleanup logs here so we
```
rartiss55, line 53: Perhaps we want a flag to disable logging for testing or something like that, but I think an additional ticket to handle that sort of thing later and saving all logs for now is fine.

kdrumm: I would agree with a flag, if we don't tie it directly to the eval parameter.  I can probably just add this now quickly.

pipeline_ray.py
kdrumm: As long as Ray doesn't have a problem pushing around the ConfigParser object (does it need to be some type of serializable?), I agree we should now be pushing that around.

ThomasSchellenbergNextCentury: If passing the configparser itself doesn't work, you can just pass the dict representation of the config: https://stackoverflow.com/questions/1773793/convert-configparser-items-to-dictionary

kdrumm:
Since this hasn't been approved yet, moving the ConfigParsing around was easier than I thought, and it was dependent on this PR anyway, I merged the config parser fix into this PR.

pipeline_ray.py
kdrumm: Forgot to explicitly say this, but I've added the flag to the mcs_config.  The parameter is 'logs_to_s3' but i'm open to renaming it.

rartiss55 (approved):

PR #26, opened 2021-07-07
Added README for creating EC2 instances.

ThomasSchellenbergNextCentury:
Documenting the steps I took to setup my EC2 instance. Hopefully it will be helpful for someone else.

PR #31, opened 2021-07-08
MCS-683 Added new ray pipeline code to generate videos.

ThomasSchellenbergNextCentury:
The goal of this ticket was to create a pipeline that generates videos of passive scenes. I did this by calling the `run_last_action.sh` script that's already in the MCS GitHub repo to run a scene and generate a corresponding video (it does so without using the `evaluation` flag), and then manually uploading the video to S3.

I copied and used Clark's latest baseline versions of the files (`autoscaler/ray_baseline_aws.yaml`, `configs/baseline_aws.ini`, `configs/mcs_config_baseline_level2.ini`, `deploy_files/baseline`) except as noted. I tested using the `run_eval` script, like:

```
./aws_scripts/run_eval.sh videos <scenes_folder> --metadata level1 --disable_validation
```

Documentation on how I made the new AMI: https://github.com/NextCenturyCorporation/mcs-pipeline/pull/26

### DIFF
--- a/autoscaler/ray_videos_aws.yaml
+++ b/autoscaler/ray_videos_aws.yaml
@@ -1,0 +1,61 @@
+#
+#   Configuration file for generating passive scene videos
+#
+# This file is a modification of: https://github.com/ray-project/ray/blob/master/python/ray/autoscaler/aws/example-full.yaml
+
+cluster_name: thomas-videos
+max_workers: 8                       # 2 is the default
+
+# Note that ray only seems to use the first availability zone, so if there are issues with that 
+# one, just need to put a different one first.
+provider:
+  type: aws
+  region: us-east-1
+  availability_zone: us-east-1a,us-east-1b,us-east-1c,us-east-1d
+  cache_stopped_nodes: True          # Default True means stopped, false means terminate.
+
+
+available_node_types:
+  ray.head.default:
+    min_workers: 0
+    max_workers: 0
+    resources: {"CPU": 4, "GPU": 1}
+    node_config:
+      InstanceType: p2.xlarge
+      ImageId: ami-0017467294d97d05c
+  ray.worker.default:
+    min_workers: 0
+    max_workers: 3
+    resources: {"CPU": 4, "GPU": 1}
+    node_config:
+      InstanceType: p2.xlarge
+      ImageId: ami-0017467294d97d05c
+      IamInstanceProfile: 
+        Arn: arn:aws:iam::795237661910:instance-profile/ray-autoscaler-worker-v1
+
+auth:
+  ssh_user: ubuntu
+
+head_node_type: ray.head.default
+
+# Copy script files to ALL nodes
+file_mounts: {
+  "/home/ubuntu/ray_script.sh": "deploy_files/videos/ray_script.sh",
+  "/home/ubuntu/check_passed_variables.sh": "deploy_files/check_passed_variables.sh",
+  "/home/ubuntu/start_x_server.sh": "deploy_files/start_x_server.sh"
+}
+
+setup_commands:
+  - pip install -U ray==1.3.0
+  - chmod +x /home/ubuntu/ray_script.sh /home/ubuntu/check_passed_variables.sh /home/ubuntu/start_x_server.sh
+  - rm -rf ~/.aws/
+
+# Command to start ray on the head node. You don't need to change this.
+head_start_ray_commands:
+  - ray stop
+  - ray start --head --port=6379 --object-manager-port=8076 --autoscaling-config=~/ray_bootstrap_config.yaml
+
+# Command to start ray on worker nodes. You don't need to change this.
+worker_start_ray_commands:
+  - ray stop
+  - ray start --address=$RAY_HEAD_IP:6379 --object-manager-port=8076
--- a/autoscaler/ray_videos_aws.yaml
+++ b/autoscaler/ray_videos_aws.yaml
@@ -3,7 +3,7 @@
 #
 # This file is a modification of: https://github.com/ray-project/ray/blob/master/python/ray/autoscaler/aws/example-full.yaml
 
-cluster_name: thomas-videos
+cluster_name: videos
 max_workers: 8                       # 2 is the default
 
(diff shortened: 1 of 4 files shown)


autoscaler/ray_videos_aws.yaml
```diff
@@ -0,0 +1,61 @@
+#
+#   Configuration file for generating passive scene videos
+#
+# This file is a modification of: https://github.com/ray-project/ray/blob/master/python/ray/autoscaler/aws/example-full.yaml
+
+cluster_name: videos
```
ThomasSchellenbergNextCentury, line 6: Here I just changed the header documentation, cluster name, and image IDs.

configs/mcs_config_videos_level1.ini
```diff
@@ -0,0 +1,24 @@
...
+
+; The S3 bucket and subfolder in which to save the video file
+; Adjust either as needed, though feel free to use the existing bucket
+s3_bucket=mcs-output-videos
+s3_folder=eval-375-test
+
+; The eval and team names are used as a prefix for the video file
+; Adjust either as needed, and either (or both) can be left blank
+evaluation_name=
+team=
```
ThomasSchellenbergNextCentury, line 24: I updated all of the properties and comments in this config file.

configs/videos_aws.ini
```diff
@@ -0,0 +1,11 @@
...
+# Execution configuration for running Ray for videos on AWS
+#
+[MCS]
+run_script = /home/ubuntu/ray_script.sh
+
+
+# Location on the head node that has the scene files and what scene files to use.
+scene_location = /home/ubuntu/scenes/tmp/
+scene_list = /home/ubuntu/scenes_single_scene.txt
+
```
ThomasSchellenbergNextCentury, line 11: I'm not sure if this config file is actually used, but I made a copy anyway.

deploy_files/videos/ray_script.sh
```diff
@@ -0,0 +1,89 @@
...
+sleep 5
+if pgrep -x Xorg > /dev/null
+then
+  echo "Killing existing X Server that started on boot"
+  sudo kill $(pgrep -x Xorg)
+  # Sleep here to ensure X Server is killed
+  sleep 5
+else
+  echo "X Server wasn't started on boot"
+fi
```
ThomasSchellenbergNextCentury, line 25: The different code in this file is here, and starting with `echo "GENERATING VIDEO: $scene_file"` below. The rationale behind this particular section is documented in the CREATE_NEW_EC2_INSTANCE_README (see branch ec2-readme).

deploy_files/videos/ray_script.sh
```diff
@@ -0,0 +1,89 @@
...
+cd $EVAL_DIR
+echo "GENERATING VIDEO: $scene_file"
+echo
+source /home/ubuntu/venv/bin/activate
+
+UNITY_APP=/home/ubuntu/unity_app/MCS-AI2-THOR-Unity-App-v0.4.3.x86_64
+
+# Read variables from MCS config file
+AWS_ACCESS_KEY_ID=$(awk -F '=' '/aws_access_key_id/ {print $2}' $mcs_configfile | xargs)
+AWS_SECRET_ACCESS_KEY=$(awk -F '=' '/aws_secret_access_key/ {print $2}' $mcs_configfile | xargs)
```
kdrumm, line 59: If the aws credentials are only needed for pushing to s3, they IAM role assigned to both head and workers _should_ provide that without leaving credentials on the disk.

ThomasSchellenbergNextCentury: Good to know! I'll try it.

ThomasSchellenbergNextCentury: It works! I've changed the code. Thanks for the heads-up!

configs/mcs_config_videos_level1.ini
```diff
@@ -0,0 +1,24 @@
...
+history_enabled=false
+; We need to save the images in order to generate the video
+save_debug_images=true
+
+; We won't use the MCS python library to upload output to S3
+evaluation=false
+
+; Set your AWS credentials!
+aws_access_key_id=
+aws_secret_access_key=
```
kdrumm, line 14: Very Minor: I don't think this affects this project, but you may want to remove these or comment them out, out of precedent.  Leaving these in, but blank would make MCS, if run, create another .aws/credentials file with blank fields.  If an .aws/credentials file exists, it takes precedence over the IAM roles and will cause issues.  Again, since I don't think this runs MCS, there is no practical need, but I thought I should mention it.

configs/videos_aws.ini
kdrumm: This file is used by our Ray code to determine where the script to run (ray_script.sh in all our cases so far) can be found.where to put scene files, which files should be run.  The run_eval.sh script does assume some of these values which is potentially problematic.  Maybe I'll create a ticket to update the run_eval.sh script to read this file.

configs/mcs_config_videos_level1.ini
ThomasSchellenbergNextCentury: You're right! They're not needed anymore, so no sense in keeping them.

kdrumm (approved):
LGTM.  It does require having the right version of scene files.  Already discussed with @ThomasSchellenbergNextCentury, but I think that issue is orthogonal to this particular PR.

cedorman:
LGTM too.

deploy_files/videos/ray_script.sh
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash
+
+# Check passed mcs_config and scene file
+source /home/ubuntu/check_passed_variables.sh
+
+EVAL_DIR=/home/ubuntu/
+SCENE_DIR="$EVAL_DIR/scenes/validation/"
+TMP_CFG_FILE="$EVAL_DIR/msc_cfg.ini.tmp"
```
rartiss55, line 8: Small typo here with **msc**_cfg, but otherwise, LGTM!

rartiss55 (approved):

PR #33, opened 2021-07-12
Mcs 767 flake8

cedorman:
No real logic changes;   make it lines so are not as long, check for True/False properly, added comments, renamed some variables.

### DIFF
--- a/aws_scripts/baseline_setup.txt
+++ b/aws_scripts/baseline_setup.txt
@@ -7,4 +7,4 @@ ray rsync_up -v autoscaler/ray_baseline_aws.yaml configs/ '~/configs/'
 ray rsync_up -v autoscaler/ray_baseline_aws.yaml scenes_single_scene.txt '~'
 ray rsync_up -v autoscaler/ray_baseline_aws.yaml <datadir> /home/ubuntu/scenes/tmp
 
-ray submit autoscaler/ray_baseline_aws.yaml pipeline_ray.py configs/opics_aws.ini configs/mcs_config_opics_oracle.ini
+ray submit autoscaler/ray_baseline_aws.yaml pipeline_ray.py configs/baseline_aws.ini configs/mcs_config_baseline_level2.ini
--- a/pipeline_ray.py
+++ b/pipeline_ray.py
@@ -38,7 +38,7 @@ def push_to_s3(source_file: pathlib, bucket: str, s3_filename: str,
     """Copy a file to a S3 bucket"""
     if client is None:
         client = boto3.client('s3')
-    logging.debug(f"Pushing {str(source_file)} to {bucket} / {s3_filename}")
+    logging.debug(f"Pushing {str(source_file)} to {bucket}/{s3_filename}")
     client.upload_file(
         str(source_file),
         bucket,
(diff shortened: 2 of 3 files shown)


ThomasSchellenbergNextCentury:
@cedorman Some questions:

- Did you just run flake8 on this specific file via the command line?
- Did you want to add a section to the README on styling and running the linter?
- Is there a plan to lint the other Python files, or is that a separate ticket?
- Is there a plan to add a linter for the shell files?
- Is there a plan to auto-lint the pipeline on commit, like we do in our other repos?

Anything that's planned for the future should be captured as separate tickets.

cedorman:
> * Did you just run flake8 on this specific file via the command line?

I ran it within PyCharm, which is my editor of choice.   

> * Did you want to add a section to the README on styling and running the linter?

The MCS DEV.md has a section on flake8.  I can copy that over 

> * Is there a plan to lint the other Python files, or is that a separate ticket?

No, the old-pipeline stuff has not been linted, and won't be.  It is supposed to go away in MCS-748, or be fixed.

> * Is there a plan to add a linter for the shell files?

No plan yet, but added a ticket (MCS-773)

> * Is there a plan to auto-lint the pipeline on commit, like we do in our other repos?

Yet another ticket!  (MCS-774)

aws_scripts/baseline_setup.txt
```diff
@@ -7,4 +7,4 @@ ray rsync_up -v autoscaler/ray_baseline_aws.yaml configs/ '~/configs/'
 ray rsync_up -v autoscaler/ray_baseline_aws.yaml scenes_single_scene.txt '~'
 ray rsync_up -v autoscaler/ray_baseline_aws.yaml <datadir> /home/ubuntu/scenes/tmp
 
-ray submit autoscaler/ray_baseline_aws.yaml pipeline_ray.py configs/opics_aws.ini configs/mcs_config_opics_oracle.ini
+ray submit autoscaler/ray_baseline_aws.yaml pipeline_ray.py configs/baseline_aws.ini configs/mcs_config_baseline_level2.ini
```
kdrumm, line 10: I know we don't want long lines, but doesn't this break the command?  I personally think with the run_eval.sh script, we should just delete these txt files anyway.

pipeline_ray.py
```diff
@@ -345,17 +424,17 @@ def parse_args():
...
     try:
         scene_runner = SceneRunner(args)
     except Exception as e:
-        logging.info(f"Exception: ", exc_info=e)
+        logging.info("Exception: ", exc_info=e)
 
     # Give it time to wrap up, produce output from the ray workers
     time.sleep(2)
     logging.info("\n*********************************")
-    input("Finished.  Hit enter to finish ")
```
kdrumm, line 361: I thought this was a ray thing.  Good idea to remove it.

kdrumm (approved):
With the new tickets, I think this looks good.  We should figure out what we want to do with the ***_setup.txt files.

ThomasSchellenbergNextCentury (approved):

aws_scripts/baseline_setup.txt
cedorman: I think that the break is not really there.   It only appears in the diff.  

I could go either way with keeping them.  I like to have a very simple set of commands that I can do manually when debugging.  Maybe move to the README?

kdrumm: You are right about the break.  Whoops, my bad.

PR #35, opened 2021-07-12
Mcs 760 read ini

kdrumm:
Added reading the locations ini file and using that in the run eval script.  Created a script that we could use in other scripts to read ini files.  We may want to move load_ini.sh at some point.


deanwetherby (approved):

ThomasSchellenbergNextCentury (changes requested):
I'm getting an error. I wrote this script:

```
#!/bin/bash
  
source aws_scripts/load_ini.sh configs/mcs_config_baseline_level2.ini

echo $MCS_team
```

But, when run, it gets the following error:

```
aws_scripts/load_ini.sh: line 23: declare: `MCS_;=Still need to address issues with how to properly store/delete these (MCS-702)': not a valid identifier
aws_scripts/load_ini.sh: line 23: declare: `MCS_;=Be sure not to leave these blank when actually running baseline': not a valid identifier
```

It looks like it can't handle the comments? (When I delete the comments, it works fine.)

kdrumm:
> I'm getting an error. I wrote this script:
> 
> ```
> #!/bin/bash
>   
> source aws_scripts/load_ini.sh configs/mcs_config_baseline_level2.ini
> 
> echo $MCS_team
> ```
> 
> But, when run, it gets the following error:
> 
> ```
> aws_scripts/load_ini.sh: line 23: declare: `MCS_;=Still need to address issues with how to properly store/delete these (MCS-702)': not a valid identifier
> aws_scripts/load_ini.sh: line 23: declare: `MCS_;=Be sure not to leave these blank when actually running baseline': not a valid identifier
> ```
> 
> It looks like it can't handle the comments? (When I delete the comments, it works fine.)

Did it fail or just print out the error?  Either way ill fix it such that it can ignore # and ; lead lines.

ThomasSchellenbergNextCentury:
@kdrumm 

> Did it fail or just print out the error? Either way ill fix it such that it can ignore # and ; lead lines.

Oh, you're right, it did work (it echoed the team). I must have missed that before. But, yeah, probably best to prevent the warnings.

kdrumm:
> @kdrumm
> 
> > Did it fail or just print out the error? Either way ill fix it such that it can ignore # and ; lead lines.
> 
> Oh, you're right, it did work (it echoed the team). I must have missed that before. But, yeah, probably best to prevent the warnings.

I added some code to skip lines starting with ; or #

ThomasSchellenbergNextCentury (approved):

PR #37, opened 2021-07-12
Mcs 762 status info

cedorman:
Print out the status during a run, showing what all the status values are for the different scenes.

### DIFF
--- a/pipeline_ray.py
+++ b/pipeline_ray.py
@@ -398,6 +398,7 @@ class SceneRunner:
             if file not in scenes_printed:
                 scenes_printed.append(file)
                 self.print_scene_status(s_status, "  ")
+        self.print_status()
 
     def print_scene_status(self, s_status, prefix=""):
         logging.info(f"{prefix}Scene: {s_status.status} - " +
--- a/pipeline_ray.py
+++ b/pipeline_ray.py
@@ -143,12 +143,12 @@ def setup_logging(log_file):
             "propagate": False
         },
         "loggers": {
-                "s3transfer": {
-                    "level": "INFO",
-                    "handlers": ["console", "log-file"],
-                    "propagate": False
-                }
-            },
+            "s3transfer": {
+                "level": "INFO",
+                "handlers": ["console", "log-file"],
+                "propagate": False
+            }
+        },
         "handlers": {
             "console": {
                 "class": "logging.StreamHandler",
@@ -184,7 +184,7 @@ def setup_logging(log_file):
 
 # Classes to keep track of status of individual scenes and (possibly
 # multiple) runs of those scenes
-class RunStatusEnum(Enum):
+class StatusEnum(Enum):
     UNKNOWN = 0,
     PENDING = 1,
     RETRYING = 2,
@@ -197,7 +197,7 @@ class RunStatusEnum(Enum):
 class RunStatus:
     exit_code: int
     output: str
-    status: RunStatusEnum
+    status: StatusEnum
     job_id = None
     retry: bool = False
 
@@ -206,7 +206,7 @@ class RunStatus:
 class SceneStatus:
     scene_file: str
     retries: int = 0
-    status: RunStatusEnum = RunStatusEnum.UNKNOWN
+    status: StatusEnum = StatusEnum.UNKNOWN
     run_statuses: List[RunStatus] = field(default_factory=list)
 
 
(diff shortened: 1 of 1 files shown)


pipeline_ray.py
```diff
@@ -183,11 +184,20 @@ def setup_logging(log_file):
 
 # Classes to keep track of status of individual scenes and (possibly
 # multiple) runs of those scenes
+class StatusEnum(Enum):
+    UNKNOWN = 0,
+    PENDING = 1,
+    RETRYING = 2,
+    SUCCESS = 3,
+    ERROR = 4,
+    ERROR_TIMEOUT = 5
```
kdrumm, line 193: I definitely like the change the enums.  I'm somewhat concerned about the ERROR_TIMEOUT value.  I think it'll work in the short term as we don't have ways to handle many different errors.

pipeline_ray.py
```diff
@@ -344,19 +348,33 @@ def run_scenes(self):
...
+        """ During the run, print out the number of completed jobs,
+        number current running, number to go, number failed, etc """
+        logging.info(f"Status for {len(self.scene_statuses)} scenes: ")
+        frequency = {}
+        current_statuses = [scene_status.status for scene_status
+                            in self.scene_statuses.values()]
+        for scene_status in current_statuses:
+            frequency[scene_status] = current_statuses.count(scene_status)
+        for key, value in frequency.items():
+            logging.info(f"    {key} -> {value}")
```
kdrumm, line 377: We should also print this status at the very end when finished.

pipeline_ray.py
cedorman: We can add more types of errors.  Or we could try to categorize them into groups, such as ERROR and ERROR_RETRYABLE.  TIMEOUT would be a retryable.

ThomasSchellenbergNextCentury, line 193: If the specific numbers don't matter, you can use `auto` here instead: https://docs.python.org/3/library/enum.html#enum.auto

Assigning specific values to Enum members implies to me that those specific values matter, but I don't think that's the case here. On the other hand, maybe I'm overthinking it :)

pipeline_ray.py
```diff
@@ -380,9 +398,10 @@ def print_results(self):
             if file not in scenes_printed:
                 scenes_printed.append(file)
                 self.print_scene_status(s_status, "  ")
+        self.print_status()
 
     def print_scene_status(self, s_status, prefix=""):
-        logging.info(f"{prefix}Scene: {s_status.status} - '+"
+        logging.info(f"{prefix}Scene: {s_status.status} - " +
                      f"'{s_status.scene_file}")
```
ThomasSchellenbergNextCentury, line 405: You technically don't need the `+` to concatenate adjacent f-strings: https://www.python.org/dev/peps/pep-0498/#concatenating-strings

pipeline_ray.py
```diff
@@ -431,6 +450,10 @@ def parse_args():
 if __name__ == '__main__':
     args = parse_args()
 
+    start_time = time.time()
+    date_str = time.strftime('%Y-%m-%d %H:%M:%S', time.localtime(start_time))
```
ThomasSchellenbergNextCentury, line 454: Not a big deal, but should we create a short function to generate the date string (or at least add a global variable for the format), so we know we're formatting the date consistently across the application?

pipeline_ray.py
cedorman: I hadn't noticed that.  I'll use auto().   Though the syntax is really ugly.

pipeline_ray.py
ThomasSchellenbergNextCentury, line 377: Using `{key}` will print the status as something like `StatusEnum.SUCCESS`, right? If you'd prefer, you can use `{key.name}` to just print `SUCCESS`

pipeline_ray.py
cedorman: Done

pipeline_ray.py
cedorman: sure, pretty easy.  Added.

pipeline_ray.py
cedorman: Nice, didn't know that either.

rartiss55:
LGTM - only thing I noticed is we may want to revisit printing each individual status at the end for long runs (maybe specific files that weren't successful would be better).

rartiss55 (approved):

ThomasSchellenbergNextCentury (approved):

PR #38, opened 2021-07-13
Mcs 770 stop restart eval

kdrumm:
Added the ability to resume a harshly stopped execution.

To test: 

1. Run an eval.  
2. Wait until at least 1 scene has completed successfully.
3. Stop the eval (ctrl + C twice works)
4. Rerun the eval with the --resume flag
5. Verify the already run scenes are not rerun.

You can also manually edit the .last_run_finished.txt file on the Ray head node.

### DIFF
--- a/pipeline_ray.py
+++ b/pipeline_ray.py
@@ -30,7 +30,8 @@ import ray
 # Number of retries before we give up on a job and call it failed
 NUM_RETRIES = 3
 
-# File the records the files that finished.  If we want to restart, we can skip these files.
+# File the records the files that finished.  If we want to restart,
+# we can skip these files.
 FINISHED_SCENES_LIST_FILENAME = "./.last_run_finished.txt"
 
 # This is logging for head node during setup and distribution.
@@ -329,15 +330,17 @@ class SceneRunner:
                 self.scene_files_list.append(base_dir / line.strip())
 
         # Filter scene_files_list based on last run file.
-        if self.resume and pathlib.Path(FINISHED_SCENES_LIST_FILENAME).exists():
+        if (self.resume and
+                pathlib.Path(FINISHED_SCENES_LIST_FILENAME).exists()):
             with open(FINISHED_SCENES_LIST_FILENAME) as last_run_list:
                 lines = last_run_list.readlines()
                 for line in lines:
-                    file=pathlib.Path(line.strip())
-                    logging.debug(f"Attempting to remove {line} from file list")
-                    if (file in self.scene_files_list ):
+                    file = pathlib.Path(line.strip())
+                    logging.debug(
+                        f"Attempting to remove {line} from file list")
+                    if (file in self.scene_files_list):
                         self.scene_files_list.remove(file)
-        
+
         self.scene_files_list.sort()
         logging.info(f"Number of scenes: {len(self.scene_files_list)}")
         logging.info(f"Scenes {self.scene_files_list}")
@@ -448,13 +451,13 @@ class SceneRunner:
         # for stop and restart
         finished_scenes_filename = pathlib.Path(FINISHED_SCENES_LIST_FILENAME)
         # missing_ok parameter was failing for me.
-        # we added self.resume check because if we are resuming, we want that previous list to stay.  
-        # Those are already completed successfully
+        # we added self.resume check because if we are resuming, we want that
+        # previous list to stay.  Those are already completed successfully
         if finished_scenes_filename.exists() and not self.resume:
             finished_scenes_filename.unlink()
         self.finished_scenes_file = open(str(finished_scenes_filename), 'a')
 
-    def on_scene_finished(self, status:SceneStatus):
+    def on_scene_finished(self, status: SceneStatus):
         self.finished_scenes_file.write(f"{status.scene_file}\n")
         self.finished_scenes_file.flush()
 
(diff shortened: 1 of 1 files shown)


pipeline_ray.py
```diff
@@ -428,6 +444,29 @@ def get_run_status(self, result: int, output: str,
...
 
+    def on_start_scenes(self):
+        # for stop and restart
+        finished_scenes_filename = pathlib.Path(FINISHED_SCENES_LIST_FILENAME)
+        print(str(finished_scenes_filename.absolute()))
+        # missing_ok parameter was failing for me.
+        # we added self.resume check because if we are resuming, we want that previous list to stay.  
+        # Those are already completed successfully
+        if finished_scenes_filename.exists() and not self.resume:
+            finished_scenes_filename.unlink()
```
kdrumm, line 455: I'm not sure why this wouldn't let me do unlink(missing_ok=True), but that failed for me.

pipeline_ray.py
```diff
@@ -250,7 +254,9 @@ def __init__(self, args):
         self.not_done_jobs = []
 
         self.get_scenes()
+        self.on_start_scenes()
```
kdrumm, line 257: just getting ready to suggest a listener on the pipeline.  Maybe, maybe not? but pipeline_ray.py is getting bigger and more complciated and may need to be refactored soon.  That isn't really related to this ticket.

These functions are here to make it easier to keep "byproduct" actions away from the code for running of the scenes.

pipeline_ray.py
```diff
@@ -428,6 +444,29 @@ def get_run_status(self, result: int, output: str,
         # Add more conditions to retry here
         return status
 
+    def on_start_scenes(self):
+        # for stop and restart
+        finished_scenes_filename = pathlib.Path(FINISHED_SCENES_LIST_FILENAME)
+        print(str(finished_scenes_filename.absolute()))
```
cedorman, line 450: Did you want this here?  it looks like a debug statement.  If you want it, use logger and add some text

pipeline_ray.py
cedorman: missing_ok was added in 3.8 so I think this is probably safer anyway.    What version of python are you on?

cedorman: Just checked:   baseline AMI is 3.7.10, so can't use missing_ok

pipeline_ray.py
kdrumm: Nope!  Good catch, I'll remove it.

pipeline_ray.py
kdrumm: Ahh, I'm using 3.8 locally, but if the AMI's are 3.7.10 that explains why I was so confused.

pipeline_ray.py
```diff
@@ -322,6 +328,16 @@ def get_scenes(self):
             if line is not None and len(line) > 0:
                 self.scene_files_list.append(base_dir / line.strip())
 
+        # Filter scene_files_list based on last run file.
+        if self.resume and pathlib.Path(FINISHED_SCENES_LIST_FILENAME).exists():
+            with open(FINISHED_SCENES_LIST_FILENAME) as last_run_list:
+                lines = last_run_list.readlines()
+                for line in lines:
+                    file=pathlib.Path(line.strip())
```
ThomasSchellenbergNextCentury, line 336: Missing the whitespace around the `=`

pipeline_ray.py
```diff
@@ -428,6 +444,29 @@ def get_run_status(self, result: int, output: str,
...
+    def on_scene_finished(self, status:SceneStatus):
+        self.finished_scenes_file.write(f"{status.scene_file}\n")
+        self.finished_scenes_file.flush()
+
+    def on_finish_scenes(self):
+        # For stop and restart
+        finished_scenes_filename = pathlib.Path(FINISHED_SCENES_LIST_FILENAME)
+        # missing_ok parameter was failing for me.
+        if finished_scenes_filename.exists():
+            finished_scenes_filename.unlink()
```
ThomasSchellenbergNextCentury, line 466: Same code as in `on_start_scenes`, right? Could the duplicated code be extracted into a separate function?

pipeline_ray.py
kdrumm: There are a couple flake8 issues that I'm fixing.

pipeline_ray.py
kdrumm: The if statement isn't exactly identical.

rartiss55:
Tested this and was able to rerun the missed scenes. Agreed about possibly coming back to this/potentially adding listeners. 

One question: do we add a scene to the resume list if there's an internal error/should we? I had a one off error that I couldn't replicate (internal to mcs python package) where some video didn't get deleted properly, and it didn't look like the scene was retried, but maybe that made sense for that case.

kdrumm:
> Tested this and was able to rerun the missed scenes. Agreed about possibly coming back to this/potentially adding listeners.
> 
> One question: do we add a scene to the resume list if there's an internal MCS error/should we? I had a one off error that I couldn't replicate (internal to mcs python package) where some video didn't get deleted properly, and it didn't look like the scene was retried, but maybe that made sense for that case.

If a scene finishes with error, but isn't supposed to retry, it will be added to the "finished" list and thus won't be run in the resume.

rartiss55 (approved):

PR #41, opened 2021-07-15
MCS-768: Add 'cora' as a valid team name

brianpippin:


brianpippin:
Leaving mit just in case we need to run any old files that have that prefix.

rartiss55 (approved):

kdrumm (approved):

PR #42, opened 2021-07-22
Mess rerun (MCS-778)

rartiss55:
Minor updates to run MESS 3.75.

### DIFF
--- a/deploy_files/mess/ray_script.sh
+++ b/deploy_files/mess/ray_script.sh
@@ -42,9 +42,10 @@ conda activate myenv
 echo Starting Evaluation:
 echo
 scene_file_basename=$(basename $scene_file)
-# TODO: Changing this manually, but we need a way to handle arguments
-# for different submissions (perhaps this is another MCS config file
-# addition)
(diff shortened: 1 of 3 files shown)


deploy_files/mess/ray_script.sh
```diff
@@ -26,12 +26,12 @@ cp $scene_file $SCENE_DIR/
 
 echo "Making temporary copy of config file ($mcs_configfile -> $TMP_CFG_FILE)"
 cp $mcs_configfile $TMP_CFG_FILE
-cp $mcs_configfile $EVAL_DIR/level1.config
+cp $mcs_configfile $EVAL_DIR/level2.config
```
kdrumm, line 29: Do we need to specify the level in the ray script?  The script should be ambiguous for whatever level it is.

rartiss55: This is actually the name of the MESS MCS config file that their code looks for (not sure why they used that name).

kdrumm (approved):

brianpippin (approved):

PR #43, opened 2021-08-17
MCS-836 - move eval code to pipeline

rartiss55:
I tested locally (see instructions below) and using a new AMI built off the MESS code with the associated branch installed

So many files were changed because I also moved the EVAL_DIR from the ray_script.sh files to the ray *.ini configs instead. Made more sense that way with these incoming changes.


configs/mcs_config_baseline_level2.ini
```diff
@@ -10,6 +10,8 @@ evaluation=false
 
 s3_bucket=evaluation-images
 s3_folder=eval-35-test
+video_enabled=true
+history_enabled=true
```
rartiss55, line 14: Since the eval/upload code is moved out of the python API, we can't rely on "evaluation" to make the videos/history, so adding these properties here. "evaluation" is more like "upload" now (perhaps we should change the name).

I think history files may get created regardless now, but I figured it didn't hurt to add that property explicitly.

README.md
```diff
@@ -142,7 +142,8 @@ Retryable: False
 
 ## Run Pipeline Locally
 
-To run the pipeline locally, make sure to update the paths in configs/test_local.ini and the EVAL_DIR in deploy_files/local/ray_script.sh to match your local machine. Then run the following:
+To run the pipeline locally, make sure to update the paths in configs/test_local.ini to match your local machine. If you'd like to test upload to S3 from a local machine, also ensure that you have your credentials and config setup correctly in ~/.aws (directions [here] (https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-quickstart.html)) and update the config in
```
rartiss55, line 145: I didn't add the aws_*key properties we had prior to the config -- IMO it makes more sense to setup the ~/.aws directory instead.

pipeline_ray.py
```diff
@@ -116,24 +118,80 @@ def run_scene(run_script, mcs_config: configparser.ConfigParser,
...
+    team = mcs_config.get("MCS", "team")
+    metadata = mcs_config.get("MCS", "metadata")
+    video_enabled = mcs_config.getboolean("MCS", "video_enabled", fallback=False)
+
+    if evaluation:
+        # find scene history file (should only be one file in directory)
+        scene_hist_matches = glob.glob(eval_dir + '/SCENE_HISTORY/' + scene_name + '*.json')
+
+        if(len(scene_hist_matches) > 0):
+            found_scene_hist = max(scene_hist_matches, key=os.path.getctime)
```
rartiss55, line 135: Getting the most recent match for history file since I don't have the timestamp appended (should only be one anyway if running on EC2).

pipeline_ray.py
```diff
@@ -116,24 +118,80 @@ def run_scene(run_script, mcs_config: configparser.ConfigParser,
...
+                [eval_name, metadata,
+                team,
+                scene_name]) + '.json'
+
+            scene_hist_file = pathlib.Path(found_scene_hist)
+
+            # update history file with additional info needed for ingest
+            logging.info("Update history file with team and eval info...")
+
+            with open(scene_hist_file, "r") as history_file:
```
rartiss55, line 146: Opening up the history file to add team/evaluation_name for ingest. This might be the best option, but there may be others (maybe we could make an additional s3 folder for performer and somehow communicate this info to ingest that way?)

pipeline_ray.py
```diff
@@ -116,24 +118,80 @@ def run_scene(run_script, mcs_config: configparser.ConfigParser,
...
+            with open(scene_hist_file, "w") as history_file:
+                json.dump(data, history_file)
+
+            # upload scene history
+            push_to_s3(scene_hist_file, bucket, scene_hist_dest, "application/json")
+        else:
+            logging.warning("History file not found for scene " + scene_name)
+
+        # find and upload videos
+        find_video_files = glob.glob(eval_dir + '/' + scene_name + '/*.mp4')
```
rartiss55, line 162: We have another ticket, MCS-730, for handling missing files more gracefully. For now, I'm simply logging a warning if there isn't a history file found or if there are NO videos found at all (nothing gets flagged really if there are some videos and not all, we simply log what we do have as its uploaded). Does that sound sufficient, or would we want to do something more?

configs/mcs_config_baseline_level2.ini
deanwetherby: Those properties feel duplicative from the settings in MCS. Do we need them at all here? Can we always assume we're uploading unless maybe the bucket is empty as a way to test?

pipeline_ray.py
deanwetherby: I don't dislike the s3 folder option here.

pipeline_ray.py
deanwetherby: It feels more unexpected and hence severe but not sure what to do about it. I guess that's what MCS-730 is for.

configs/mcs_config_baseline_level2.ini
rartiss55: To clarify, you mean dropping "evaluation" and just using the s3_bucket/s3_folder to upload, or are there other properties you think can go?

deanwetherby: As we continue to move to a more generic pipeline (i.e. not so "evaluation" centered), we could replace `evaluation_name` with `job_name` and `eval_dir` with `working_dir` or something like that. Then the underlying runner code can decide what, if anything, to do with those properties.

pipeline_ray.py
ThomasSchellenbergNextCentury: I agree with Dean that other options (like using S3 folders) are worth investigating. The scene history files can get fairly large, and if we're loading and rewriting each one, that could cause a non-trivial performance decrease. Additionally, this could break if the scene history file's format ever changes.

configs/mcs_config_baseline_level2.ini
rartiss55: Made another ticket for this + other things to figure out with our config file: https://nextcentury.atlassian.net/browse/MCS-867

pipeline_ray.py
rartiss55: Moved these things back to the python API for now and put this in a separate ticket as something to investigate: https://nextcentury.atlassian.net/browse/MCS-867

pipeline_ray.py
rartiss55: Closed MCS-730 since the original issue seems resolved.

pipeline_ray.py
```diff
@@ -116,24 +118,73 @@ def run_scene(run_script, mcs_config: configparser.ConfigParser,
...
+    video_enabled = mcs_config.getboolean("MCS", "video_enabled", fallback=False)
+
+    if evaluation:
+        # find scene history file (should only be one file in directory)
+        scene_hist_matches = glob.glob(eval_dir + '/SCENE_HISTORY/' + scene_name + '*.json')
+
+        if(len(scene_hist_matches) > 0):
+            found_scene_hist = max(scene_hist_matches, key=os.path.getctime)
+            hist_filename_no_ext = os.path.splitext(found_scene_hist)[0]
+            timestamp = hist_filename_no_ext[-15:]
```
rartiss55, line 137: Grabbing the timestamp from the filename to make sure I'm grabbing the history files + videos from the same run.

README.md
```diff
@@ -38,9 +38,12 @@ and UI related functionality will work correctly (these can be turned off for te
 - **metadata** - has to be either `level1`, `level2`, or `oracle`
 - **evaluation_name** - has to be one of the following, in the exact format: `eval_3-75`, `eval_4`, `eval_5`, `eval_6`, `eval_7`, `eval_8`
 - **evaluation** - must be set to `true`
+- **history_enabled** - must be set to `true`
 - **team** - has to be either `mess`, `cora` (or `mit` if rerunning a pre-3.75 eval set), `opics`, or `baseline`
 - **s3_bucket** - should be `evaluation-images`
-- **s3_folder** - has to be the folder we store output for the current eval (right now, `eval-3.75`)
+- **s3_folder** - json output - has to be the folder we store output for the current eval (right now, `eval-3.75`)
+- **s3_movies_folder**: required (value should be `eval-resources-4` for eval 4) - only mp4 output, MediaConvert will copy all mp4s to the "s3_folder" config property as well
```
brianpippin, line 45: s3_movies_folder should be 'raw-eval-4/' and 'eval-resources-4' should be what the s3_folder should be, "raw-eval-#" is where the movies go to then be converted and then moved to 'eval-resources-4'

configs/opics_aws.ini
```diff
@@ -8,4 +8,4 @@ run_script = /home/ubuntu/ray_script.sh
 # Location on the head node that has the scene files and what scene files to use.
 scene_location = /home/ubuntu/scenes/tmp/
 scene_list = /home/ubuntu/scenes_single_scene.txt
-
+eval_dir = /home/ubuntu/mcs_eval3-3.5.0
```
brianpippin, line 11: Will this change in the future depending on the eval?  Would be nice if we could get them or us to make it generic.

README.md
rartiss55: Thanks! Updated.

configs/opics_aws.ini
rartiss55: Currently, yes. I'm not sure if we can get them to stick to one directory but its a good idea.

pipeline_ray.py
brianpippin: I'm confused by this code.  The history files no longer have timestamps on them, we wanted to get rid of the scene history files having timestamp, so any newer ones should overwrite that, unless we changed something to add timestamps back in?

rartiss55: We have timestamps appended to the filenames when they are written locally, but not when they are uploaded to S3. If running an eval, we should be clearing out old history files anyway, but this doesn't happen if testing locally/not using the run_eval script, so I thought it'd be best just in case to sort by most recent + grab the timestamp from the filename to make sure we have the correct history file/it matches the timestamp on the video files.

brianpippin: So long as the timestamped files don't get uploaded to S3 I'm cool with that :)

brianpippin (approved):

PR #44, opened 2021-10-04
[DRAFT] Ongoing updates to video generation pipeline. (Updating to v0.4.5)

ThomasSchellenbergNextCentury:


scenes_single_scene.txt
```diff
@@ -1,3 +0,0 @@
-juliett_0028_07.json
-juliett_0028_09.json
-juliett_0028_08.json
```
ThomasSchellenbergNextCentury, line 3: I don't think these scene names need to be hard-coded here?

rartiss55: You're right -- this was leftover before the code to auto create this file was added. Might be useful to keep it around as an example and commit a test scene to the repo or something, but otherwise I don't think its needed. I added it to a tech debt ticket.

PR #51, opened 2021-11-03
MCS-954 Changes for running the Eval 4 NYU baselines.

ThomasSchellenbergNextCentury:


deploy_files/baseline/ray_script.sh
```diff
@@ -9,6 +9,21 @@ SCENE_DIR="$eval_dir/scenes/"
...
+sleep 5
+if pgrep -x Xorg > /dev/null
+then
+  echo "Killing existing X Server that started on boot"
+  sudo kill $(pgrep -x Xorg)
+  # Sleep here to ensure X Server is killed
+  sleep 5
+else
+  echo "X Server wasn't started on boot"
+fi
```
ThomasSchellenbergNextCentury, line 25: Copied from the video generation pipeline, which were both built off a normal Ubuntu Server AMI, and this was necessary to make xorg/unity work consistently.

autoscaler/ray_baseline_aws.yaml
```diff
@@ -51,6 +51,7 @@ setup_commands:
   - pip install --upgrade pip
   - pip install -U "ray[default]"==1.3.0
   - pip install aioredis==1.3.1
+  - pip install aiohttp
```
ThomasSchellenbergNextCentury, line 54: This was a failed attempt to solve an ongoing issue -- see the discussion in the ta2-api Slack channel.

deploy_files/baseline/ray_script.sh
deanwetherby: It's interesting you have to do all this differently and it seems to be based on the AMI. I've only ever used the DL AMI for MCS so the busid is always 0:30:0. Maybe using a different AMI is causing more grief than it should be.

ThomasSchellenbergNextCentury: > the busid is always 0:30:0

Let me clarify what's happening here:
1. Set the busid to a wrong number (I used 0:31:00 here)
2. Kill xorg. It will automatically try to immediately restart itself, but fail because of the bad busid.
3. Set the busid to the correct number (0:30:00)
4. Start xorg manually using the `:0` flag, which we know from experience will work correctly.

Note that steps 3 and 4 both happen in the `start_x_server.sh` script, called here on line 28.

> Maybe using a different AMI is causing more grief than it should be

Maybe, though I had already figured out this solution for the video generation pipeline, and it's relatively straightforward (if a bit wonky)... maybe it's just hindsight? We can consider that if we ever need to rebuild any of these AMIs from scratch, I think.

autoscaler/ray_baseline_aws.yaml
```diff
@@ -51,6 +51,7 @@ setup_commands:
   - pip install --upgrade pip
   - pip install -U "ray[default]"==1.3.0
   - pip install aioredis==1.3.1
+  - pip install aiohttp==3.7.4.post0
```
ThomasSchellenbergNextCentury, line 54: Got it working, with Kyle's help. Without installing this specific version of this package, ray spits out an endless stream of error messages while it's running... though it's only happening on the baseline and video pipelines (maybe just because of how I built the original AMIs?).

ThomasSchellenbergNextCentury:
This PR is ready/waiting for review, in case it wasn't apparent before. If the NYU team has any further changes to the baseline code, I can just add it as a separate PR.

kdrumm (approved):
Barring any better solution to the mentioned X server issues, this looks good to me.

PR #52, opened 2021-11-04
Eval4 opics

kdrumm:
This has a couple components that could have been separate:

- Added documentation for timestamping log lines
- Updated opics execution (AMI, run script, setup commands, config files)
- Added timeout to opics' script via sed
- Added flags to customize cluster name and number of workers in a temporary config file
- Added failed status if no history file.
- Added log parser script for getting rough metrics from log files.  This file is fairly rough around the edges.

### DIFF
--- a/autoscaler/ray_opics_aws.yaml
+++ b/autoscaler/ray_opics_aws.yaml
@@ -4,8 +4,8 @@
 #
 # This file is a modification of: https://github.com/ray-project/ray/blob/master/python/ray/autoscaler/aws/example-full.yaml
 
-cluster_name: opics-kdrumm-rerun
-max_workers: 10                         # 2 is the default
+cluster_name: opics
+max_workers: 8                         # 2 is the default
 
 # Note that ray only seems to use the first availability zone, so if there are issues with that 
 # one, just need to put a different one first.
--- a/autoscaler/ray_opics_aws.yaml
+++ b/autoscaler/ray_opics_aws.yaml
@@ -4,15 +4,15 @@
 #
 # This file is a modification of: https://github.com/ray-project/ray/blob/master/python/ray/autoscaler/aws/example-full.yaml
 
-cluster_name: opics
-max_workers: 8                       # 2 is the default
+cluster_name: opics-kd
+max_workers: 25                       # 2 is the default
 
 # Note that ray only seems to use the first availability zone, so if there are issues with that 
 # one, just need to put a different one first.
 provider:
   type: aws
   region: us-east-1
-  availability_zone: us-east-1a,us-east-1b,us-east-1c,us-east-1d
+  availability_zone: us-east-1a,us-east-1b,us-east-1c,us-east-1d,us-east-1e,us-east-1f
   cache_stopped_nodes: True          # Default True means stopped, false means terminate.
 
 
@@ -23,13 +23,13 @@ available_node_types:
     resources: {"CPU": 4, "GPU": 1}
     node_config:
       InstanceType: p2.xlarge
-      ImageId: ami-08b3c4d663c483fd0
+      ImageId: ami-0b152d8074b893b46
   ray.worker.default:
     min_workers: 0
     resources: {"CPU": 4, "GPU": 1}
     node_config:
       InstanceType: p2.xlarge
-      ImageId: ami-08b3c4d663c483fd0
+      ImageId: ami-0b152d8074b893b46
       IamInstanceProfile: 
         Arn: arn:aws:iam::795237661910:instance-profile/ray-autoscaler-worker-v1
 
@@ -48,6 +48,7 @@ file_mounts: {
 
 setup_commands:
   - sudo cp /home/ubuntu/asound.conf /etc/asound.conf
+  - pip install --upgrade pip
   - pip install -U "ray[default]"==1.3.0
   - rm -rf ~/.aws/
   - pip install aioredis==1.3.1
--- a/autoscaler/ray_opics_aws.yaml
+++ b/autoscaler/ray_opics_aws.yaml
@@ -23,13 +23,13 @@ available_node_types:
     resources: {"CPU": 4, "GPU": 1}
     node_config:
       InstanceType: p2.xlarge
-      ImageId: ami-0d238012b47fe6ad9
+      ImageId: ami-01d14613654e5ea30
   ray.worker.default:
     min_workers: 0
     resources: {"CPU": 4, "GPU": 1}
     node_config:
       InstanceType: p2.xlarge
-      ImageId: ami-0d238012b47fe6ad9
+      ImageId: ami-01d14613654e5ea30
       IamInstanceProfile: 
         Arn: arn:aws:iam::795237661910:instance-profile/ray-autoscaler-worker-v1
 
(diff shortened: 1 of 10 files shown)


pipeline/log_parser.py
```diff
@@ -0,0 +1,167 @@
...
+        while True:
+            rem = str(ts % 60)
+            rem = f"0{rem}" if len(rem) == 1 else rem
+            ts = int(ts / 60)
+            delim = ":" if ts > 0 else ""
+            result = f"{delim}{rem}{result}"
+            if ts < 1:
+                return result
+
+    def print(self):
```
deanwetherby, line 89: Print function here reminds me of the dunder (double underscore) class methods `__str__` and `__repr__` which could be optionally used for this class. Then you could just `print(exestats)` to get the same result.

pipeline/log_parser.py
```diff
@@ -0,0 +1,167 @@
...
+    split_log('tmp2.txt')
+
+
+if __name__ == '__main__':
+    # parser = argparse.ArgumentParser(
+    #    description='Parse log files.'
+    # )
+    
+    # args = parser.parse_args()
+    main(None)
```
ThomasSchellenbergNextCentury, line 167: I'm trying to figure out what this file is for. I'm assuming it's a standalone script, since it's not called anywhere else. Is the docstring under `main` correct?

kdrumm: This file is the rough script that searches logs and parses out timing information.  Its also currently broken because I fixed opics entry script which changes some of the logs.  I'm not sure yet what to do with this.

autoscaler/ray_opics_aws.yaml
```diff
@@ -48,9 +48,11 @@ file_mounts: {
 
 setup_commands:
   - sudo cp /home/ubuntu/asound.conf /etc/asound.conf
+  - pip install --upgrade pip
   - pip install -U "ray[default]"==1.3.0
   - rm -rf ~/.aws/
   - pip install aioredis==1.3.1
+  - pip install shapely==1.7.1
```
ThomasSchellenbergNextCentury, line 55: Should we add a comment here about this fixing a bug in the Eval 4 submission, in case it causes a problem in a future submission and we need to modify it?

aws_scripts/run_eval.sh
```diff
@@ -47,14 +39,42 @@ while [ $# -gt 0 ]; do
...
+    NEW_CFG=".ray-configs/${CFG_FILE%.yaml}-$CLUSTER_SUFFIX.yaml"
+    # cp "$RAY_CONFIG" "$NEW_CFG"
+    CL=`grep cluster_name: $RAY_CONFIG`
+    NEW_CLUSTER_LINE="$CL-$CLUSTER_SUFFIX"
+    sed "s/cluster_name:\s*\S*/$NEW_CLUSTER_LINE/g" $RAY_CONFIG > $NEW_CFG
+    RAY_CONFIG=$NEW_CFG
+
+    echo "Replaced cluster name line with:"
+    echo "  `grep cluster_name: $NEW_CFG`"
+fi
```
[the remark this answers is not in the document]
kdrumm: This adds a suffix to the clustername in a new config file.  This allows you to run multiple clusters with the same yaml (essentially).  For example, i plan on using this to run a cluster for alpha-level2 and another cluster for bravo-level2 etc.  I'm adding more documentation right now.

autoscaler/ray_opics_aws.yaml
kdrumm: sure

pipeline_ray.py
```diff
@@ -613,6 +618,11 @@ def get_run_status(
             logging.info(f"Timeout occured for file={scene_file_path}")
             status.retry |= True
             status.status = StatusEnum.ERROR_TIMEOUT
+        if not reported_success:
+            logging.info(
+                f"Pipeline reported failure for file ={scene_file_path}")
+            status.retry |= True
+            status.status = StatusEnum.ERROR
```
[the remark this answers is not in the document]
kdrumm: That's a good idea.  I'm not sure if we have access to the IP where this is logged though.   I remember it being tricky finding some things with ray.  Usually there will be a log with an IP just above this so finding it isn't hard.  I'll see if I can add it here though.

autoscaler/ray_opics_aws.yaml
```diff
@@ -4,15 +4,15 @@
 #
 # This file is a modification of: https://github.com/ray-project/ray/blob/master/python/ray/autoscaler/aws/example-full.yaml
 
-cluster_name: opics
-max_workers: 8                       # 2 is the default
+cluster_name: opics-kdrumm
```
ThomasSchellenbergNextCentury, line 7: I'm not sure if we want to keep your name in here.

aws_scripts/run_eval.sh
```diff
@@ -5,6 +5,24 @@
...
+#  metadata_level: metadata level to run MCS at
+#  resume: Signals the head node to attempt to resume its last run.  This just removes
+#    files that the head node recorded as completed in its last run.  
+#  cluster_suffix: Adds a suffix to the cluster name already in the ray configuration
+#    yaml file.  A new copy of the configuration will be put in .ray_configs/.  This 
+#    is useful to run multiple clusters using the same yaml file.  For example, running
+#    a bunch of clusters for different sets of files and/or metadata levels.
+#  workers: override number of workers
+#  disable_validation: Disables the validation on the MCS configuration.  Useful when
+#    testing or running against the development infrastructure.
```
kdrumm, line 25: We should consider if some of this in a more robust script (maybe python?) with templates etc.

pipeline/log_parser.py
```diff
@@ -0,0 +1,262 @@
...
+            'Upload', "Pushing .*\\.log"),
+    ]
+
+
+def get_default_matchers():
+    return [
+        LogMatcher(
+            'Start', "In run scene."),
+        LogMatcher('Initialization', "Found path:"),
+        LogMatcher('Finish scene', "History file timestamp")
```
rartiss55, line 206: May be worth adding the line here for upload as well, that should be the same for all the performers.

pipeline/log_parser.py
```diff
@@ -0,0 +1,262 @@
...
+
+if __name__ == '__main__':
+
+    parser = argparse.ArgumentParser(
+        description='Parse log files from run_eval.sh.  Logs must include timestamps from the ts command.'
+    )
+    parser.add_argument('--output-dir', '-o', default='./split_logs/')
+    parser.add_argument('log_file')
+    parser.add_argument(
+        '--matchers', '-m', choices=['OPICS', 'MESS', 'CORA', 'BASELINE'], default='OPICS')
```
rartiss55, line 259: Should the default be blank?

pipeline/log_parser.py
kdrumm: The history file is uploaded directly after the last one here.  The main reason I chose what I did instead was it was easier to match.

pipeline/log_parser.py
kdrumm: I made the matcher group default blank and the field required.  I also added help text.

rartiss55 (approved):

PR #53, opened 2021-11-05
Eval 4 - MESS runner code

rartiss55:
Still may need to finalize the AMI used, but otherwise ready for review.


pipeline_ray.py
```diff
@@ -216,8 +219,18 @@ def run_scene(
             + ".txt"
         )
 
+        log_s3_filename_latest = (
```
rartiss55, line 222: This was to get a consistently named log file for displaying in the UI. I will likely take this out, since I think Brian had a better idea with adding all the timestamped logs to history records using boto on the ingest side (or we can even just use boto to update what we need after the fact).

ThomasSchellenbergNextCentury (approved):

kdrumm (approved):
Seems reasonable to me so far

PR #54, opened 2021-11-11
Topdown Video Generation Pipeline

ThomasSchellenbergNextCentury:
The topdown video generation pipeline is just a tweaked version of the existing video generation pipeline. I used this pipeline for generating the topdown videos for the passive agent scenes. I'm not sure how useful this pipeline will be in the future, but I figured I should commit my code anyway, just in case.

For comparison:
- A few lines in `deploy_files/topdown/ray_script.py` are different than the corresponding video pipeline file.
- A few properties in `configs/mcs_config_topdown_level1.ini` are different as well.

The rest should be pretty much the same.

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,7 +279,9 @@ we're going to use mess_tasks_runner.py as if it were an interactive tool.
     * Make a file with the correct list of tasks and set the TASK_FILE_PATH to point to it
     * Run the tasks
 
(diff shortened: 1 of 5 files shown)


kdrumm (approved):
I didn't run it, but it seems reasonable.

PR #57, opened 2021-12-08
Mcs 1026 config template

kdrumm:
Includes a lot of new features:
- New python run_eval.py script
- Using mako templating engine
- Use single templates for mcs config and ray config files
- Create general stats and timing estimates
- Orchestrate variable number of clusters
- Orchestrate custom permutations of scene folders and metadata levels

This incorporates tickets: 744, 1027, 1026

Questions/Issues: 
- Terminology and naming - I am unsure how to name things an it is evident in the code
- Documentation - documentation is messy and probably not clear
- Setting variable and varsets - Is this clear? easy? necessary?  I like the idea of hierarchical variable sets, but is there a more intuitive way?
- .tmp_pipeline_ray folder accumulates folders with each run and requires a manual clear

More work:
- Merge CORA PR when it comes in and validate (will be in this PR prior to final merge)
- Better error capturing and output parsing (MCS-1049)
- Better status and timing (MCS-1049)
- Json or yaml status capture that is rerun-able (MCS-1049)


aws_scripts/run_eval.py
```diff
@@ -0,0 +1,518 @@
...
+import threading
+import queue
+import time
+import random
+
+# TODO:
+# update after all teams eval4 is merged
+# make sure parameters get pass through (disable_validation, dev, resume, etc)
+#   update dev
+# Add status
```
ThomasSchellenbergNextCentury, line 23: Making note of this TODO so we don't forget it.

aws_scripts/run_eval.py
```diff
@@ -0,0 +1,518 @@
...
+    run_eval(varset, local_scene_dir,
+             metadata=metadata, disable_validation=True)
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description="Run multiple eval sets containing scenes using ray.  There are two modes.  " +
+        "One uses the '--config_file' option.  The other uses the '--local_scene_dir' option.")
+    parser.add_argument(
+        "--config_file",
```
ThomasSchellenbergNextCentury, line 418: Would it make sense to add some single letter variations for a few of these arguments? Like `-c` for `--config_file`, `-n` for `--num_clusters`, `-m` for `--metadata`, etc.?

aws_scripts/run_eval.py
```diff
@@ -0,0 +1,518 @@
...
+        varset: ['opics', 'kdrumm']
+    eval-groups:
+        - metadata: ['level1', 'level2']
+          parent-dir: 'mako-test/parent'
+        - metadata: ['level2', 'oracle']
+          dirs: ['mako-test/dirs/dir1', 'mako-test/dirs/dir2']
+          
+    This example will use the 'opics.yaml' and 'kdrumm.yaml' files in the 'variables' directory to fill the templates.
+    It will level1 and level2 test using
+    """
```
ThomasSchellenbergNextCentury, line 504: Good documentation. Could we also add it (or part of it) to the README?

aws_scripts/run_eval.py
```diff
@@ -0,0 +1,518 @@
...
+
+
+def single_test():
+
+    varset = ['opics', 'kdrumm']
+    metadata = 'level2'
+    local_scene_dir = 'eval4/single'
+
+    run_eval(varset, local_scene_dir,
+             metadata=metadata, disable_validation=True)
```
ThomasSchellenbergNextCentury, line 410: Are these two functions still used?

aws_scripts/run_eval.py
```diff
@@ -0,0 +1,518 @@
...
+        metadata_list = get_array(group, my_base, 'metadata')
+        for metadata in metadata_list:
+            parents = get_array(group, my_base, 'parent-dir')
+            dirs = get_array(group, my_base, 'dirs')
+            for dir in dirs:
+                my_override = {}
+                log_dir = dir.split("/")[-1]
+                my_override["log_name"] = f"{log_dir}-{metadata}.log"
+                evals.append(EvalParams(
+                    varset, dir, metadata, override=my_override))
```
ThomasSchellenbergNextCentury, line 379: Could these lines and the similar lines in `create_eval_set_from_folder` be extracted into a separate shared function?

aws_scripts/run_eval.py
```diff
@@ -0,0 +1,518 @@
...
+        params = " ".join(args)
+        execute_shell(
+            f"ray submit {self._config_file.as_posix()} {file} {params}",
+            self._log_file)
+
+
+def run_eval(varset, local_scene_dir, metadata="level2", disable_validation=False,
+             dev_validation=False, resume=False, override_params={},
+             log_file=None, cluster="", output_logs=False) -> pathlib.Path:
+    """Runs an eval and returns the ray config file as a pathlib.Path object."""
```
ThomasSchellenbergNextCentury, line 184: Does this do any sort of time logging for us? If not, could it? Or is this not the right place for that behavior? I know Dean was keen on us doing it at some point.

aws_scripts/run_eval.py
```diff
@@ -0,0 +1,518 @@
...
+    ray_cfg_file.write_text(ray_cfg)
+
+    # Generate MCS config
+    mcs_cfg_template = Template(
+        filename='mako/templates/mcs_config_template.ini')
+    mcs_cfg = mcs_cfg_template.render(**vars)
+    mcs_cfg_file = working / f"mcs_config_{team}_{metadata}.ini"
+    mcs_cfg_file.write_text(mcs_cfg)
+
+    # mcs_cfg_file = f"configs/mcs_config_{team}_{metadata}.ini"
```
ThomasSchellenbergNextCentury, line 222: Extra line?

aws_scripts/run_eval.py
```diff
@@ -0,0 +1,518 @@
...
+    vars = {**vars, **override_params}
+    vars['metadata'] = metadata
+
+    # Setup Tail
+    if log_file and output_logs:
+        lt = LogTailer(log_file, f"c{cluster}: ")
+        lt.tail_non_blocking()
+
+    # Setup working directory
+    now = datetime.now().strftime('%Y%m%d-%H%M%S')
```
ThomasSchellenbergNextCentury, line 196: Should the date/time format string be a global variable? We use it in `main` as well, and might need to add more in the future.

aws_scripts/run_eval.py
ThomasSchellenbergNextCentury: Oh wait is it this line?
```
print(all_status.time_string())
```

aws_scripts/run_eval.py
```diff
@@ -0,0 +1,518 @@
...
+    # if we want to capture the logging.
+    cmd = f"unbuffer {cmd} 2>&1 | ts -S"
+    if (log_file):
+        with open(log_file, "a") as f:
+            subprocess.run([cmd, "|", "ts"], stdout=f,
+                           stderr=subprocess.STDOUT, shell=True)
+    else:
+        subprocess.run(cmd, shell=True)
+
+    # os.system(cmd)
```
ThomasSchellenbergNextCentury, line 148: Still needed?

mako/eval.yaml
```diff
@@ -0,0 +1,7 @@
+base:
+  varset: ['opics', 'kdrumm']
+eval-groups:
+  - metadata: ['level1', 'level2']
+    parent-dir: 'mako-test/parent'
+  - metadata: ['level2', 'oracle']
+    dirs: ['mako-test/dirs/dir1', 'mako-test/dirs/dir2']
```
ThomasSchellenbergNextCentury, line 7: Is this supposed to be a runnable example? If so, should we have a few scenes committed in this repo under the `mako-test/parent` and `mako-test/dirs` folders? (Or have instructions in the README on how to setup the scenes to run this example.)

mako/variables/kdrumm.yaml
```diff
@@ -0,0 +1 @@
+clusterUser: -kdrumm
```
ThomasSchellenbergNextCentury, line 1: How do we feel about each person having their own variable file committed to the repo? I don't feel strongly either way.

aws_scripts/run_eval.py
ThomasSchellenbergNextCentury: Though I'll also note that it's unfinished (but I think you know that already).

mako/variables/kdrumm.yaml
kdrumm: This is probably a bad idea in the end.  But I do find it useful to tag the clusters with our names.  We could either have a user file that isn't checked in or some other mechanic.

aws_scripts/run_eval.py
kdrumm: There is some timing and other ticket to improve it later.  I'll also note here that I'm not fond of the function name 'time_string'.

I think the time_string just tells you how much time is left.  It should probably say how much time has elapsed as well (if it doesn't)

mako/eval.yaml
kdrumm: This is a good question.  It is a runnable example, but i didn't check in the files.  I'm open to ideas on what we actually want to check in.

aws_scripts/run_eval.py
kdrumm: I also need to add to this to make the template/variable/etc location configurable.  I also suspect naming the base directory 'mako' is not what we want in the end.

aws_scripts/run_eval.py
kdrumm: yes, in my next push.  Thoughts on using -v for varsets when -v usually means verbose?  (do we want to rename varsets?)

aws_scripts/run_eval.py
```diff
@@ -0,0 +1,519 @@
...
+
+import pathlib
+import yaml
+import os
+import threading
+import queue
+import time
+
+# TODO:
+# update after CORA eval4 is merged
```
ThomasSchellenbergNextCentury, line 19: Reminder

README.md
```diff
@@ -75,8 +75,59 @@ In order to test the pipeline and evaluations, the following is helpful:
...
+        varset: ['opics', 'personal']
+    eval-groups:
+        - metadata: ['level1', 'level2']
+          parent-dir: 'mako-test/parent'
+        - metadata: ['level2', 'oracle']
+          dirs: ['mako-test/dirs/dir1', 'mako-test/dirs/dir2']
+          
+    This example will use the 'opics.yaml' and 'personal.yaml' files in the 'variables' directory to fill the templates.
+    It expects to have a directory 'mako-test/parent/' which has one or more subdirectories filled with scenes.  It also
+    expects the following directories with scene files: 'mako-test/dirs/dir1', 'mako-test/dirs/dir2'.
```
ThomasSchellenbergNextCentury, line 125: Does this need to be in both the README and the script? Seems harder to maintain that way.

aws_scripts/run_eval.py
kdrumm: todo has one item left which references another ticket.

cedorman:
I'm confused about what is still needed and what the preferred way of running things is.  There was pipeline_ray.py, but then a script run_eval.sh to set things up for it;  we are transitioning to run_eval.py, which is fine, but if it is the preferred way of running an eval, then it should be clear.  At this point, there are too many ways of doing it. 
- I think that run_eval.py should be in the top level, rather than aws_scripts; remove run_eval.sh
- configs/*.ini files should go away?   hard to know what is still needed and not needed if they are still there
- what is still needed in aws_scripts?   
- does pipeline_ray.py need to be at the top level?   Will people be running that, or is it something that is now a sub-process to run_eval.py?  
- it's confusing to have have both a config file and a command-line parser; I would vote to just have a config file.

mako/templates/ray_template_aws.yaml
```diff
@@ -0,0 +1,69 @@
...
+max_workers: ${workers}                       # 2 is the default
+
+idle_timeout_minutes: ${idle_timeout_minutes}
+
+# Note that ray only seems to use the first availability zone, so if there are issues with that 
+# one, just need to put a different one first.
+provider:
+  type: aws
+  region: ${region}
+  availability_zone: us-east-1a,us-east-1b,us-east-1c,us-east-1d,us-east-1e,us-east-1f
```
rartiss55, line 17: availability_zones probably need to be "${region}a, ${region}b, ${region}c, etc" as well

mako/eval.yaml
rartiss55: Maybe we could check in 2-3 intphys style files from a validation dataset and use those?

rartiss55:
I need more time to digest/try out run_eval.py, but I like the templating and wanted to leave a few thoughts regarding some of the points mentioned at the top:
- Regarding documentation: I think this probably needs its own ticket. There is a LOT of information here + in confluence regarding Ray/the pipeline/how we run the evals. I do think nailing down how exactly we want to run everything like Clark was saying will help with this a little bit.
- For cleaning things up after a run, like in .tmp_pipeline_ray: would it make sense to have a line that cleans up files of a certain type/pattern after the 'ray down' call in ray_eval.py? Maybe something else to clean up everything once all jobs have finished?

kdrumm:
> I'm confused about what is still needed and what the preferred way of running things is. There was pipeline_ray.py, but then a script run_eval.sh to set things up for it; we are transitioning to run_eval.py, which is fine, but if it is the preferred way of running an eval, then it should be clear. At this point, there are too many ways of doing it.
> 
> * I think that run_eval.py should be in the top level, rather than aws_scripts; remove run_eval.sh
> * configs/*.ini files should go away?   hard to know what is still needed and not needed if they are still there
>   -- Nope, they are still needed;  used in run_eval, line 206
> * what is still needed in aws_scripts?
> * does pipeline_ray.py need to be at the top level?   Will people be running that, or is it something that is now a sub-process to run_eval.py?
> * it's confusing to have have both a config file and a command-line parser; I would vote to just have a config file.

For your points:

- Moved the run_eval.py script around and removed a bunch of stuff in autoscaler.  I need to check and see if this will work with the video creation configs.
- The {team}-aws.ini are still around.  They would probably not be hard to remove, but i'd prefer to leave that for future work.
- aws_scripts is gone
- moved pipeline_ray into a ray_script folder
- We should talk more about the config file and command line.  The goal here was the config is "what to run" and the command line was "how to run it".  The goal was to allow us to share the config with minimal changes.  This does get confused a little with the idea of name tagging clusters, number of clusters and number of workers.  I also added the capability from the command line parser to run a single run kind of like the old shell script.  I should probably move that to another python file entirely.

kdrumm:
> I need more time to digest/try out run_eval.py, but I like the templating and wanted to leave a few thoughts regarding some of the points mentioned at the top:
> 
> * Regarding documentation: I think this probably needs its own ticket. There is a LOT of information here + in confluence regarding Ray/the pipeline/how we run the evals. I do think nailing down how exactly we want to run everything like Clark was saying will help with this a little bit.
> * For cleaning things up after a run, like in .tmp_pipeline_ray: would it make sense to have a line that cleans up files of a certain type/pattern after the 'ray down' call in ray_eval.py? Maybe something else to clean up everything once all jobs have finished?

The only reason I don't want to clean up the files after a run is that is where I am saving the resume configuration file in the next ticket as well as you can inspect the yaml to make sure it is correct.  It should probably be renamed if it isn't going to be that temporary.  Thoughts?

deanwetherby (approved):
This is a big step in the right direction. I need to get a better handle on the mako config templating as it seems to be pivotal for future pipeline tasks. There were some references in the README to autoscaler which appears to be gone. As others have mentioned, we'll need to unify on the approach and ensure that it is well documented.

PR #59, opened 2021-12-21
Mcs 1049 detailed status

kdrumm:


ray_scripts/pipeline_ray.py
```diff
@@ -279,13 +279,13 @@ def setup_logging(log_file):
...
 # Classes to keep track of status of individual scenes and (possibly
 # multiple) runs of those scenes
-class StatusEnum(Enum):
-    UNKNOWN = auto()
-    PENDING = auto()
-    RETRYING = auto()
-    SUCCESS = auto()
-    ERROR = auto()
-    ERROR_TIMEOUT = auto()
+class StatusEnum(str, Enum):
```
deanwetherby, line 282: Is there a reason for multiple inheritance here?

kdrumm: This allows you to set strings to the keys in the enum.

ray_scripts/pipeline_ray.py
```diff
@@ -279,13 +279,13 @@ def setup_logging(log_file):
...
+    UNKNOWN = 'UNKOWN'
+    PENDING = 'PENDING'
+    RETRYING = 'RETRYING'
+    SUCCESS = 'SUCCESS'
+    ERROR = 'ERROR'
+    ERROR_TIMEOUT = 'ERROR_TIMEOUT'
```
deanwetherby, line 288: This is an interesting pattern with python enums that I've seen in other places (ENUM_THING = 'ENUM_THING'). You can get the enum as a string using name as in `StatusEnum.UNKNOWN.name`. If that works, then auto could suffice as the enum values.

ray_scripts/pipeline_ray.py
```diff
@@ -573,6 +575,30 @@ def print_status(self):
...
             logging.info(f"    {key.name} -> {value}")
 
+    def output_status(self):
+        json_status = {'Succeeded': 0, 'Failed': 0,
+                       'Total': len(self.scene_files_list)}
+        statuses = []
+        json_status['statuses'] = statuses
+
+        for key in self.scene_statuses:
+            value = self.scene_statuses[key]
```
deanwetherby, line 585: Is scene_statuses a dictionary? if so, you can iterate over the key AND value

```
for key, value in self.scene_statuses.items():
  print(f"{key}-{value}")
```

run_eval.py
```diff
@@ -27,64 +29,155 @@
...
+            # for some reason deep copy fixes a yaml issue
+            'varset': copy.deepcopy(self.varset),
+            'metadata': [self.metadata],
+            'dirs': [self.scene_dir],
+        }
+
+        if self.override:
+            d['override'] = self.override
+        if self.file_names:
+            d['files'] = self.file_names
```
deanwetherby, line 95: Can override and files just go in the dictionary creation conditionally? Or does it cause problems if the key exists with some None or empty list value?
```
d = {
  'varset': copy.deepcopy(self.varset),
  'metadata':[self.metadata],
  'dirs': [self.scene_dir],
  'override': self.override if self.override else None,
  'files': self.file_names if self.file_names else None # or else []
}
```

deanwetherby (approved):
Picked on some very minor things but nothing that will cause a problem with runtime.

ray_scripts/pipeline_ray.py
kdrumm: I'm not exactly sure why, but using StatusEnum.UNKNOWN.name didn't work.  This is inside the pipeline_ray.py code that is run remotely so its harder to debug and takes a while to run tests unfortunately.  I'm still doing some tests, but soon I might revert it to the multiple inheritance.

run_eval.py
kdrumm: This should work fine.  Sourcery recommended using or instead of the if then statement so I'm using that.

ray_scripts/pipeline_ray.py
kdrumm: I think in my earlier testing, i missed the .items().  Since i don't need the key in this case, this made me realize I could probably use values().

run_eval.py
```diff
@@ -82,19 +82,15 @@ def get_resume_eval_params(self):
...
         """Returns dictionary to write as yaml"""
-        d = {
-            # for some reason deep copy fixes a yaml issue
+        # for some reason deep copy fixes a yaml issue
+        return {
             'varset': copy.deepcopy(self.varset),
             'metadata': [self.metadata],
             'dirs': [self.scene_dir],
+            'override': self.override or {},
+            'files': self.file_names or []
```
deanwetherby, line 91: very nice

brianpippin (approved):
The code looks clean to me.

PR #61, opened 2021-12-23
minor pipeline updates

kdrumm:
Changes:

- Added `user` to automatic varset if the file exists.  In the standard location, this file is gitignored.  This allows each developer to have their own user file without interfering with each other and checking into git.
- Updated documentation
- Added constants for period for status print and resume file generation
- Removed support for passing the resume flag to `pipeline_ray.py`.  Resume is now handled outside of ray.
- Added support for `number of workers` and `cluster user` to command line arguments.  This further supports the idea that the config is supposed to be 'what is run' while the command line arguments are more 'how it is run'
- Added argument for `team` in `run_eval_single.py`


deanwetherby (approved):

rartiss55:
Haven't tried it out, but code LGTM.

rartiss55 (approved):

PR #63, opened 2022-06-13
Added some fixes for library updates

kdrumm:
The pipeline was broken when Jacob and I tried to test it with opics.  The main fix was to just using the latest version of ray 1 (1.13).  This change should be noted and kept in mind for all performers.

Note: The old version seemed to work for Thomas with videos, so I'm not sure why it didn't work for OPICS.

### DIFF
--- a/mako/templates/ray_template_aws.yaml
+++ b/mako/templates/ray_template_aws.yaml
@@ -52,10 +52,11 @@ file_mounts: {
 setup_commands:
   - sudo cp /home/ubuntu/asound.conf /etc/asound.conf
   - pip install --upgrade pip
-  - pip install -U "ray[default]"==1.3.0
-  - pip install aioredis==1.3.1
-  - pip install aiohttp==3.7.4.post0 # Is this ok for all performers?
-  - rm -rf ~/.aws/
+#  - aws configure set region us-east-1
+  - pip install -U "ray[default]"==1.13.0
+#  - pip install aioredis==1.3.1
+#  - pip install aiohttp==3.7.4.post0 # Is this ok for all performers?
+#  - rm -rf ~/.aws/
   ${additional_setup_commands}
 
 # Command to start ray on the head node. You don't need to change this.
--- a/mako/templates/ray_template_aws.yaml
+++ b/mako/templates/ray_template_aws.yaml
@@ -52,7 +52,6 @@ file_mounts: {
 setup_commands:
   - sudo cp /home/ubuntu/asound.conf /etc/asound.conf
   - pip install --upgrade pip
-#  - aws configure set region us-east-1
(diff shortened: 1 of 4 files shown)


mako/templates/ray_template_aws.yaml
```diff
@@ -52,10 +52,10 @@ file_mounts: {
...
   - sudo cp /home/ubuntu/asound.conf /etc/asound.conf
   - pip install --upgrade pip
-  - pip install -U "ray[default]"==1.3.0
-  - pip install aioredis==1.3.1
-  - pip install aiohttp==3.7.4.post0 # Is this ok for all performers?
-  - rm -rf ~/.aws/
+  - pip install -U "ray[default]"==1.13.0
+#  - pip install aioredis==1.3.1
+#  - pip install aiohttp==3.7.4.post0 # Is this ok for all performers?
+#  - rm -rf ~/.aws/
```
ThomasSchellenbergNextCentury, line 58: If we're saving these lines for some reason, we should leave a comment why.

ThomasSchellenbergNextCentury:
@kdrumm Should we also update the `ray` version in our pip requirements file? Or does that version not matter?

kdrumm:
> @kdrumm Should we also update the `ray` version in our pip requirements file? Or does that version not matter?

I didn't and it worked.  Its something that should also probably be tested.

rartiss55:
> The pipeline was broken when Jacob and I tried to test it with opics. The main fix was to just using the latest version of ray 1 (1.13). This change should be noted and kept in mind for all performers.
> 
> Note: The old version seemed to work for Thomas with videos, so I'm not sure why it didn't work for OPICS.

The older version worked for MESS previously last week, so I'm not sure why either. Either way, tested it out and it seems to work for OPICS and MESS as well. 

Looked through the change logs for Ray. The only two items I noticed that jumped out at me were:

1.9: Ray Clusters on AWS are colocated in one Availability Zone to reduce costs & latency (19051)
1.11: Ray no longer starts Redis by default. Cluster metadata previously stored in Redis is stored in the GCS now. (sounds like redis is no longer a requirement)

It seems like the clusters thing was more specific to spot instances, and I think the 1.11 item is worth noting, but I thought I'd post them for sanity checking.

rartiss55:
> > @kdrumm Should we also update the `ray` version in our pip requirements file? Or does that version not matter?
> 
> I didn't and it worked. Its something that should also probably be tested.

For me, upgrading to 1.13 locally resulted in some trouble with file mounts/errors on starting my cluster. It looks like the fullpath (not relative) is needed in the *aws.yaml config for those.

brianpippin (approved):

rartiss55 (approved):

PR #64, opened 2022-06-20
MCS-1362: Changes for Final CORA run

brianpippin:
Just putting this out there to make sure I am not doing anything egregious while running the pipeline.

### DIFF
--- a/mako/eval.yaml
+++ b/mako/eval.yaml
@@ -1,8 +1,11 @@
 base:
   varset: ['cora']
 eval-groups:
+  # - metadata: ['level2']
+  #   dirs: ['../oscar_containers', '../papa_obstacles', '../quebec_occluders',
+  #   '../romeo_object_permence', '../sierra_agent_identification', '../tango_moving_target',
+  #   '../uniform_holes', '../victor_lava','../whiskey_ramps', '../xray_solidity', '../zulu_support_relations',
+  #   '../omega_tool_use','../yankee_support_relations']
   - metadata: ['level2']
-    dirs: ['../oscar_containers', '../papa_obstacles', '../quebec_occluders',
-    '../romeo_object_permence', '../sierra_agent_identification', '../tango_moving_target',
-    '../uniform_holes', '../victor_lava','../whiskey_ramps', '../xray_solidity', '../zulu_support_relations',
-    '../omega_tool_use','../yankee_support_relations']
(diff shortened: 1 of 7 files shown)


mako/variables/cora.yaml
```diff
@@ -1,4 +1,4 @@
-ami: ami-00da8aa4b07be3e9a
-cluster: "cora"
+ami: ami-03d77a16ead707e25
+cluster: "cora-brian-test"
```
kdrumm, line 2: There is another way to do this without putting your name in a file that gets checked in.  You can create a user.yaml file in the variables directory.  Inside, if you put `clusterUser: -brian`, it'll add that to the end of your cluster.

brianpippin: Thanks Kyle, this is exactly why I posted my first run, so that I could improve and figure out what I didn't do quite right.

mako/templates/ray_template_aws.yaml
```diff
@@ -52,7 +52,7 @@ file_mounts: {
 setup_commands:
   - sudo cp /home/ubuntu/asound.conf /etc/asound.conf
   - pip install --upgrade pip
-  - pip install -U "ray[default]"==1.3.0
+  - pip install -U "ray[default]"==1.13.0
   - pip install aioredis==1.3.1
   - pip install aiohttp==3.7.4.post0 # Is this ok for all performers?
```
rartiss55, line 57: I think you should be able to get rid of these dependencies (PR here https://github.com/NextCenturyCorporation/mcs-pipeline/pull/63), but let me know if that doesn't work for you.

ThomasSchellenbergNextCentury (approved):

mako/eval.yaml
```diff
@@ -1,7 +1,15 @@
 base:
-  varset: ['opics']
+  varset: ['cora']
```
rartiss55, line 2: We should probably revert all our various versions for these now that we're done with the eval (perhaps we can rename the file too to eval_sample.yaml or something after everything's merged).

mako/variables/default.yaml
```diff
@@ -1,4 +1,4 @@
-workers: 8
+workers: 75
```
rartiss55, line 1: I think this should be reverted as well just so nothing makes it to the prod bucket during any future testing.

brianpippin:
@rartiss55 - made the requested changes, will wait to merge mine once some of these other pipeline changes get merged.

rartiss55 (approved):

PR #66, opened 2022-06-21
Opics configurations for eval 5

jaudick:


mako/variables/default.yaml
```diff
@@ -1,21 +1,18 @@
 workers: 8
 region: us-east-1
-clusterUser: ""
+clusterUser: "jaud"
```
ThomasSchellenbergNextCentury, line 3: I think this needs to be reset before we merge this branch.

ThomasSchellenbergNextCentury: Also can't you use the `-u` command line argument for this? Or am I mistaken?

kdrumm: I noticed this and thought i'd chime in.  There is a gitignored file that is also read for configuration and the `clusterUser` is typically put in there so it stays only for the users machine.  I forget what the file name is, but i know its hidden in the docs somewhere (probably user.yaml or something like that).  @ThomasSchellenbergNextCentury 's idea is also sound if one prefers that method.

jaudick:
This should be ready to merge now

mako/eval.yaml
```diff
@@ -1,7 +1,5 @@
 base:
   varset: ['opics']
 eval-groups:
-  - metadata: ['level1', 'level2']
-    parent-dir: 'mako-test/parent'
-  - metadata: ['level2', 'oracle']
-    dirs: ['mako-test/dirs/dir1', 'mako-test/dirs/dir2']
+  - metadata: ['level2']
+    dirs: ['']
```
rartiss55, line 5: I think we should keep the old version to illustrate how different metadata levels/directory properties work if needed.

PR #68, opened 2022-06-27
MCS-1347 - MESS g4dn changes

rartiss55:
Changes to get g4dn.xlarge machines working -- main things:
- deploy_files/start_x_server.sh changes since the graphics card is a bit different (just removing `--use-display-device=None` and adding `--no-use-display-device`)
- *.yaml template file updates so that we can use slightly different machine types/resources if needed

~~In addition, you will likely need to comment out the `"UseDisplayDevice" "None"` option out of the /etc/X11/xorg.conf file (I did this and re-snapshotted, but we could also commit an xorg.conf + copy it up on initialization, unless there's something I'm missing).~~ No longer needed with Brian's update/`--no-use-display-device` flag.

May be taken care of depending on how you do the previous step, but **COPY YOUR AMI FIRST.** I tried setting some of this up again against one of my p2xlarge snapshots and got a warning about deleting my existing volume.

### DIFF
--- a/configs/mess_aws.ini
+++ b/configs/mess_aws.ini
@@ -6,4 +6,4 @@ run_script = /home/ubuntu/ray_script.sh
 
 scene_location = /home/ubuntu/scenes/tmp/
 scene_list = /home/ubuntu/scenes_single_scene.txt
-eval_dir = /home/ubuntu/mess_eval4
+eval_dir = /home/ubuntu/mess_eval5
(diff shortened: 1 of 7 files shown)


ThomasSchellenbergNextCentury (approved):

rartiss55:
Closing for now -- referenced this PR in the JIRA ticket in case we return to it in the future.

PR #69, opened 2022-07-15
MCS-1385 NYU Red Team Pipeline Eval 5

ThomasSchellenbergNextCentury:
- Made separate `configs/*_aws.ini` and `mako/variables/*.yaml` files for each NYU-supported task. Removed the corresponding "baseline" files.
- Kept the `deploy_files/baseline/` folder, because it's associated with the "baseline" team name, and I didn't want to change the team name, because I'm assuming it would break the UI. Made updates as needed.
- Added a new `filePrefix` mako config variable that can be used instead of the configured "team" for identifying the pipeline's config file paths.

TODO:
- Update the docs for running NYU.

Design Notes:
- The NYU code has 8 different scripts (in their own folders) that are each intended to be run on a specific type of task. Combining them into a single folder would take some refactoring (there are a lot of overlapping filenames and hard-coded file paths), and so would writing a single python script that imports the other scripts in order to call them (they're all basically just "main" functions). Furthermore, our pipeline's "eval_dir" needs to point to the folder for the specific task, because that's where the scene history and video output are saved. Therefore the best approach seemed to be creating separate configs (each with a unique "eval_dir") for each NYU task.

Open Questions:
- Splitting the NYU pipeline into separate configs for each of their supported tasks was less invasive to the NYU code, but added a lot of files to the repository. Is that OK? At least they're all named "nyu_*" so it's fairly obvious.


deploy_files/baseline/ray_script.sh
```diff
@@ -31,24 +26,13 @@ fi
 source /home/ubuntu/start_x_server.sh
 
 # Clear out the directories
-echo Clearing History at "$eval_dir"/SCENE_HISTORY/
 rm -f "$eval_dir"/SCENE_HISTORY/*
-echo Clearing "$SCENE_DIR"
```
rartiss55, line 36: Had to double take at this but at this point, I suppose this step isn't entirely necessary (although I don't think it hurts to leave it in for the other performers).

mako/nyu_eval_5.yaml
```diff
@@ -0,0 +1,19 @@
+base:
```
rartiss55, line 1: Normally I wouldn't necessarily want to commit this back into main, but since the baseline case is so unique with how its all split into different directories, may be worthwhile.

rartiss55:
Looks like logs/empty.txt may have been committed by mistake?

rartiss55:
> Open Questions:
> 
> * Splitting the NYU pipeline into separate configs for each of their supported tasks was less invasive to the NYU code, but added a lot of files to the repository. Is that OK? At least they're all named "nyu_*" so it's fairly obvious.

I'm fine with it, and thanks for the explanation given in the "Design Notes" -- its intuitive enough but may be worth leaving an additional note somewhere (in the files themselves or the readme maybe?) as to why that is to remind us in ~7 months when we look at this again :)

mako/nyu_eval_5.yaml
ThomasSchellenbergNextCentury: Yeah, it seemed like a weird enough case to have a separate example config. Open to suggestions on ways to improve it.

ThomasSchellenbergNextCentury:
@rartiss55 

> Looks like logs/empty.txt may have been committed by mistake?

I was getting an error for not having a `logs/` directory (though now, of course, I can't remember how to reproduce it, ha) and wasn't sure the best way to address it. Ensuring the repo would have a `logs/` directory when you cloned it was the solution I had implemented.

rartiss55:
> @rartiss55
> 
> > Looks like logs/empty.txt may have been committed by mistake?
> 
> I was getting an error for not having a `logs/` directory (though now, of course, I can't remember how to reproduce it, ha) and wasn't sure the best way to address it. Ensuring the repo would have a `logs/` directory when you cloned it was the solution I had implemented.

@ThomasSchellenbergNextCentury I had forgotten about this too and had trouble reproducing it (hah!) turns out it was for run_eval_single.py -- I made a separate PR to address it, let me know what you think: https://github.com/NextCenturyCorporation/mcs-pipeline/pull/74

ThomasSchellenbergNextCentury:
@rartiss55 Looks great. I went ahead and removed the "empty" file.

ThomasSchellenbergNextCentury:
Going to merge this PR by the end of the day.

PR #75, opened 2023-03-09
Updates in preparation for eval 6

rartiss55:
Updating some of the "current eval" variables from eval 5 to 6.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,8 @@ Config File API (yaml):
 This script assumes you have the ts.  you can install ts via `sudo apt install moreutils`
 This script assumes you have the unbuffer.  you can install ts via `sudo apt install expect`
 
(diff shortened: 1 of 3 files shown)


phile-caci (approved):
Looks good, I searched through the code and only found eval-5 references in those three docs too.

rartiss55:
Wanted to add a couple things for anyone testing prior to next week:
- I was able to get a cluster going, but since my performer AMI from eval 5 is using a very old version of MCS, I wasn't able to run any scenes through.
- The folders referenced are for when we officially run validation + the eval. Please use different test S3 folders if you want to do some of your own testing before hand, as any history files uploaded to the 'eval-6' buckets will be uploaded (or attempt to be uploaded) to the database. 
- If you want to double check all of your configs for doing a run, you can use the `--dry_run` flag (you'll likely need `--disable_validation` as well).

PR #79, opened 2023-03-22
Use non-standard team names like opics-v2

ThomasSchellenbergNextCentury:
In `mako/variables/<team>.yaml`, if you want to change the `team` name to something different like `opics-v2`, you can now add the `filePrefix` setting (in the same file) using the original team name (like `filePrefix: opics`), which will use the proper files in the `configs` and `deploy_files` folders. (The "filePrefix" already existed, and worked for the `configs` folder, but not the `deploy_files`, thus this PR.)


rartiss55 (approved):

phile-caci (approved):
Looks good to me.

PR #81, opened 2023-04-28
Cora eval 6 (Ready for review)

phile-caci:
Changes for CORA

### DIFF
--- a/deploy_files/cora/ray_script.sh
+++ b/deploy_files/cora/ray_script.sh
@@ -7,24 +7,86 @@ source /home/ubuntu/check_passed_variables.sh
 # shellcheck disable=SC2154
 echo "Running CORA with config $mcs_configfile and scene $scene_file using eval dir $eval_dir"
 
-# Start X
-source /home/ubuntu/start_x_server.sh
-export DISPLAY=:0
-
 # Clear out directories
 echo Clearing History at "$eval_dir"/SCENE_HISTORY/
 rm -f "$eval_dir"/SCENE_HISTORY/*
-
 export MCS_CONFIG_FILE_PATH=$mcs_configfile
 
 # Run the Performer code
-echo Starting Evaluation:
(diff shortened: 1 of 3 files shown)


ray_scripts/pipeline_ray.py
```diff
@@ -123,7 +123,7 @@ def run_scene(
         cmd.split(" "), stdout=subprocess.PIPE, stderr=subprocess.STDOUT
     )
     lines = []
-    for line in io.TextIOWrapper(proc.stdout, encoding="utf-8"):
+    for line in io.TextIOWrapper(proc.stdout, encoding="utf-8", errors='ignore'):
```
phile-caci, line 126: CORA was outputting some unicode, which raised an error here.  Added 'ignore' to ignore encoding issues.

deploy_files/cora/ray_script.sh
```diff
@@ -7,24 +7,86 @@ source /home/ubuntu/check_passed_variables.sh
...
 rm -f "$eval_dir"/SCENE_HISTORY/*
-
 export MCS_CONFIG_FILE_PATH=$mcs_configfile
 
 # Run the Performer code
-echo Starting Evaluation:
+echo Starting Evaluation for CORA:
 
 # Adjust for where they hardcoded the scene file to be read from, might be different next collab/evaluation run
-cd "$eval_dir"/assets/scene_jsons/test_scenes_small || exit
```
rartiss55, line 24: You might know this, but for future reference, the $eval_dir property is in cora_aws.ini and this is the location that gets referenced by pipeline_ray.py (looks like the value is the same as what's used at the end of this file here too, /home/ubuntu/CoraAgent).

rartiss55 (approved):

jaudick (approved):

ThomasSchellenbergNextCentury (approved):

PR #82, opened 2023-05-03
MCS-1783 Remove local running code

jaudick:
According to the Ticket : _"This was more useful in the early days where we were still building the pipeline, not as much now IMO."_


phile-caci (approved):
Looks good to me!

rartiss55:
I think you can also get rid of deploy_files/local/ray_script.sh as part of this ticket.

rartiss55:
> @rartiss55 Just pushed a commit

Thanks! I made another pass, sorry I realized I missed one more file (configs/local.yaml). That should be it though!

rartiss55 (approved):

PR #83, opened 2023-10-02
MCS-1771: Monitor process script to handle timeouts

rartiss55:
Will be off by default, but wanted to get this in since it happened for me for a couple of scene types during Eval 6. 

Basically, if a timeout situation occurs and the user stops taking any steps: there is code on the Python API side to create a scene history + video files if there is inactivity for x amount of time, but something else needs to handle actually quitting TA1's runner code + exiting Unity, otherwise things just hang.


deploy_files/mess/monitor_process.py
```diff
@@ -0,0 +1,123 @@
...
+import psutil
+import logging
+
+logger = logging.getLogger()
+logging.basicConfig(level=logging.INFO, stream=sys.stdout)
+
+
+def main(scene_file_basename, eval_dir):
+    """
+    This script is only needed for situations where the performer code gets
```
rartiss55, line 16: Hopefully this is enough detail in case this needs to be incorporated for other performers.

ThomasSchellenbergNextCentury: This documentation is great! I'll leave myself a ticket to implement this in the OPICS script (and @phile-caci might want to do the same for CORA), but will probably wait to focus on it until closer to our validation runs.

mako/variables/mess.yaml
```diff
@@ -4,3 +4,7 @@ region: us-east-2
 ami: ami-0642c523a480d0ae1 # east-2 region ami for 'mess-eval6-agency-with-patch'
 #ami: ami-0d3bcdaa29a0c659f # east-2 region ami for 'mess-eval6-final-v4' (pre-agency patch)
 cache_stopped_nodes: False
+# Properties needed in case of timeouts
+additional_file_mounts: '"/home/ubuntu/monitor_process.py": "deploy_files/mess/monitor_process.py",'
```
ThomasSchellenbergNextCentury, line 8: Should we move `monitor_process.py` up one folder (so it's just in `deploy_files/`)?

deploy_files/mess/monitor_process.py
```diff
@@ -0,0 +1,123 @@
...
+    deploy_files/mess/ray_script.sh for an example).
+    """
+    logging.info("monitor_process.py: starting with the following args: ")
+    logging.info(
+        f"monitor_process.py: scene_file_basename: {scene_file_basename}"
+    )
+    logging.info(f"monitor_process.py: eval_dir: {eval_dir}")
+    time.sleep(10)
+    # TA1 run command and full scene file path
+    full_cmd = "python src/script_mess_clean.py scenes/" + scene_file_basename
```
ThomasSchellenbergNextCentury, line 42: Is there any way to maybe pass this in as a command line argument so this script's not MESS-specific?

mako/variables/mess.yaml
```diff
@@ -4,3 +4,7 @@ region: us-east-2
...
+additional_setup_commands: "- python3 -m pip install psutil==5.9.5"
+has_monitor_process: false
```
ThomasSchellenbergNextCentury, line 10: Any reason we don't want this to default to true?

deploy_files/mess/monitor_process.py
```diff
@@ -0,0 +1,123 @@
...
+    has_monitor_process is set to 'true'.
+    Also be sure to update ray_script.sh with the relevant parts (see
+    deploy_files/mess/ray_script.sh for an example).
+    """
+    logging.info("monitor_process.py: starting with the following args: ")
+    logging.info(
+        f"monitor_process.py: scene_file_basename: {scene_file_basename}"
+    )
+    logging.info(f"monitor_process.py: eval_dir: {eval_dir}")
+    time.sleep(10)
```
ThomasSchellenbergNextCentury, line 40: What's up with all the 10 sec sleeps?

deploy_files/mess/monitor_process.py
```diff
@@ -0,0 +1,123 @@
...
+
+            with open(full_scene_file_path, "r") as f:
+                data = json.load(f)
+                proper_filename = data["name"]
+
+            history_files = f"{eval_dir}/SCENE_HISTORY/{proper_filename}*.json"
+            scene_hist_matches = glob.glob(history_files)
+
+            # check if history file exists
+            if len(scene_hist_matches) > 0:
```
ThomasSchellenbergNextCentury, line 69: Will this have trouble stopping if the TA1 code errors out and doesn't generate a scene history file? Is that important?

deploy_files/mess/monitor_process.py
rartiss55: I tried making this more generic but ran into issues (having to update the pgrep command since the TA1 run command would be in two process names if it were an arg, which I could get around, but then ran into the argument limit on EC2). I think its probably best to leave this as is for now.

mako/variables/mess.yaml
rartiss55: See comment below re: making this script more generic.

deploy_files/mess/monitor_process.py
rartiss55: Its my understanding that just because you send a signal to terminate a process doesn't mean it terminates immediately, so I wanted to give some extra time. I think this case here though was trying to give time for the TA1 run command to actually start.

mako/variables/mess.yaml
rartiss55: Yes -- adding those sleeps I mentioned below adds a bit of overhead that I didn't feel like dealing with with our potential machine limitations, and this timeout situation (should be) pretty atypical, so I planned on leaving it off until it was necessary.

deploy_files/mess/monitor_process.py
rartiss55: We've done different things to try to generate a history file if there's an exception or timeout, so that shouldn't happen, but its possible that is still some weird edge case somehow.

deploy_files/mess/monitor_process.py
ThomasSchellenbergNextCentury: OK, got it. Thanks.

mako/variables/mess.yaml
ThomasSchellenbergNextCentury: Understood. Thanks.

deploy_files/mess/monitor_process.py
ThomasSchellenbergNextCentury: Cool. Thanks.

rartiss55:
> The code looks good. Is there anything you want us to do in order to test it ourselves? I haven't run the MESS code before, so getting setup might be more trouble than it's worth. I'm fine with approving the PR with just looking over the code though :)

I think that's fine -- I'll run an additional retest now just in case that I've reverted some things but I don't expect anyone else to get spun up on how MESS worked in Eval 6. I'll leave this open for a bit though before merging in case others want to take a look.

rartiss55:
Finished retest mentioned in earlier comment, things look good. I don't expect anyone to do this, but in case anyone wants to test with MESS: 

- be sure to update mess.yaml with local variables (i.e. eval-35-test for the s3_folder and s3_movie_folder, updating cluster and user names, updating `timeout` to a smaller interval so you don't have to wait around as long, setting `has_monitor_process` to true.)
- set `workers` in mess.yaml to 0 so you're only running a head node + if you're running on one node its easier to guarantee that things are recovering properly between scenes
- make a mako/*.yaml run file referencing a directory with one or a few scenes to test (some scenes that triggered the timeout situation for MESS include 'kilo_0011_01' and 'kilo_0011_02')

phile-caci (approved):
LGTM!

PR #87, opened 2023-11-16
[DRAFT] MESS Eval 7

rartiss55:


ray_scripts/pipeline_ray.py
```diff
@@ -27,6 +27,11 @@
 from typing import List
 
 import boto3
+
+os.environ["RAY_DEDUP_LOGS"] = "0"
```
rartiss55, line 31: Deduping the logs (at least for me) removed stuff that wasn't really duplicates, but related to another job, so I am disabling it. The line following was really helpful with debugging the object_store issue, so I left it in but commented it out (basically when you run "ray memory" on the head node, it gives you sources/line numbers for any object store items).

ray_scripts/pipeline_ray.py
```diff
@@ -512,17 +520,25 @@ def run_scenes(self):
...
-        self.not_done_jobs = job_ids
-        while self.not_done_jobs:
-            done_jobs, self.not_done_jobs = ray.wait(self.not_done_jobs)
-            for done_ref in done_jobs:
-                result, output, success = ray.get(done_ref)
-                scene_status = self.scene_statuses.get(done_ref)
+                self.incomplete_jobs.append(job_id)
+
+        # delete reference to last job_id
+        del job_id
```
rartiss55, line 533: The last reference from the above for loop was being saved, so I'm deleting it.

ray_scripts/pipeline_ray.py
```diff
@@ -548,6 +563,15 @@ def run_scenes(self):
...
                 self.output_status()
 
+                # Delete the ray reference to this finished job. This will
+                # allow ray to stop the worker if it's now idle. Sometimes
+                # a short sleep is needed after deleting the reference.
+                self.jobs_to_scenes.pop(finished_job_id)
+                del finished_job_id
+
+            # delete reference to old finished jobs list
+            del finished_jobs
```
rartiss55, line 573: Deleting the finished_jobs list once we're done iterating on it (it should only be one item at a time anyway with how ray.wait() works on default). Basically ray.wait() won't return a new item until one is actually completed, so if you only have one job that completes and the rest hang, that item might persist in the object store/a machine may not idle when it should. You can add a timeout to ray.wait() to help mitigate this (and the latest docs now recommend to do so), but I did not experiment with that.

ray_scripts/pipeline_ray.py
```diff
@@ -512,17 +520,25 @@ def run_scenes(self):
...
-            for done_ref in done_jobs:
-                result, output, success = ray.get(done_ref)
-                scene_status = self.scene_statuses.get(done_ref)
+                self.incomplete_jobs.append(job_id)
+
+        # delete reference to last job_id
+        del job_id
+
+        while self.incomplete_jobs:
+            finished_jobs, self.incomplete_jobs = ray.wait(self.incomplete_jobs)
```
rartiss55, line 536: Copied Thomas' changes over, but tweaked things a bit. Besides what's mentioned elsewhere, I removed the unfinished_jobs variable (am just using self.incomplete_jobs) and am not attempting to iterate on it to clean up any references (I didn't think it was necessary/the cause of any object store issues based on my testing).

ray_scripts/pipeline_ray.py
rartiss55: Also, I don't know why this is the best way to do ray environment variables, but trying to set them in the start up commands in ray_template_aws.yaml (which I think is recommended?) didn't work for me.

run_eval.py
```diff
@@ -312,7 +312,7 @@ def __init__(
         dry_run=False,
         base_dir="mako",
         group_working_dir=RAY_WORKING_DIR,
-        num_retries: int = 3,
+        num_retries: int = 1,
```
rartiss55, line 315: I was wondering if retries was messing with my statuses when I was troubleshooting (I don't think they were) but I'm leaving this as 1 for my run because generally I've found that 3 retries is pointless.

mako/templates/ray_template_aws.yaml
```diff
@@ -22,13 +22,13 @@ available_node_types:
...
     max_workers: 0
-    resources: {"CPU": 4, "GPU": ${1 if utilize_head_node_for_work is True else 0}}       # Setting GPU to 0 won't run work on the head node
+    resources: {"CPU": 8, "GPU": ${1 if utilize_head_node_for_work is True else 0}}       # Setting GPU to 0 won't run work on the head node
     node_config:
       InstanceType: p3.2xlarge
       ImageId: ${ami}
   ray.worker.default:
     min_workers: 0
-    resources: {"CPU": 4, "GPU": 1}
+    resources: {"CPU": 8, "GPU": 1}
```
ThomasSchellenbergNextCentury, line 31: Should I be doing this with OPICS too?

rartiss55: It might for logging accuracy, but I'm pretty sure for running things it only matters that the GPU info is correct anyway.

mako/variables/mess.yaml
```diff
@@ -1,10 +1,14 @@
...
-#ami: ami-0d3bcdaa29a0c659f # east-2 region ami for 'mess-eval6-final-v4' (pre-agency patch)
+ami: ami-0b9e2f88ec61e1b6c # east-2 region ami for 'mess-evaluation-7-submission'
+# ami: ami-0ed5aa45a1bd9db99 # this is the east-2 ami for 'mess-eval-7-agency-fix'
+# ami: ami-0a4152550d29b5655 # this is the east-2 ami for 'mess-eval7-tools-fix'
 cache_stopped_nodes: False
+utilize_head_node_for_work: False
 # Properties needed in case of timeouts
 additional_file_mounts: '"/home/ubuntu/monitor_process.py": "deploy_files/mess/monitor_process.py",'
-additional_setup_commands: "- python3 -m pip install psutil==5.9.5"
+additional_setup_commands: "- pip install pydantic==1.10.7"
```
rartiss55, line 12: Needed this pydantic version for the MESS instance specifically (some other dependency we use messes with this) -- moved the psutil install to ray_template_aws.yaml since it was for the monitor process anyway.

mako/variables/mess.yaml
```diff
@@ -1,10 +1,14 @@
 cluster: "mess"
+workers: 10
 team: "mess"
 region: us-east-2
-ami: ami-0642c523a480d0ae1 # east-2 region ami for 'mess-eval6-agency-with-patch'
-#ami: ami-0d3bcdaa29a0c659f # east-2 region ami for 'mess-eval6-final-v4' (pre-agency patch)
+ami: ami-0b9e2f88ec61e1b6c # east-2 region ami for 'mess-evaluation-7-submission'
```
rartiss55, line 5: In case I have to drop off, please check the AMI before running anything outstanding:
- mess-evaluation-7-submission: original submission
- mess-eval-7-agency-fix: fix from Dec 6th running agency to mitigate some of the rate limiting
- mess-eval7-tools-fix: includes dev build patch for tool scenes with RotateObject hotfixes.

mako/templates/ray_template_aws.yaml
```diff
@@ -22,13 +22,13 @@ available_node_types:
   ray.head.default:
     min_workers: 0
     max_workers: 0
-    resources: {"CPU": 4, "GPU": ${1 if utilize_head_node_for_work is True else 0}}       # Setting GPU to 0 won't run work on the head node
+    resources: {"CPU": ${num_cpus}, "GPU": ${1 if utilize_head_node_for_work is True else 0}}       # Setting GPU to 0 won't run work on the head node
```
rartiss55, line 25: Added num_cpus as a property since I recall Thomas having trouble setting it to 8 (although I'm not sure why).

rartiss55:
Ready for review, but leaving as an open PR in case there are any reruns/issues with little divergences between branches.

